PR #28, opened 2020-01-03
Jbs plugin paths

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/28)


jsarabia:
This PR modifies the plugins to allow for a list of possible binary paths for each platform, rather than just one. The first binary that both exists and is executable is chosen.

PR #33, opened 2020-01-27
RBA refactor

aunger:
Refactor a few things, with no change in functionality:
- Use fewer database row ints as function parameters in favor of more type-safe objects
- Move ContentType helper collections into ContentType.kt
- Split long functions
- Rename a few vals and functions for clarity
- Add a little kdoc

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/33)


jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/persistence/repositories/CollectionRepository.kt
```diff
@@ -312,6 +246,80 @@ class CollectionRepository(
...
+        dsl: DSLContext,
+        source: Collection,
+        language: Language
+    ): ResourceMetadataEntity {
+        // Check for existing resource containers
+        val existingMetadata = metadataDao.fetchAll(dsl)
+        val matches = existingMetadata.filter {
+            it.identifier == source.resourceContainer?.identifier &&
+                    it.languageFk == language.id &&
+                    it.creator == "Orature" &&
```
jsarabia: Assuming we could "resume" working on a resource container that wasn't created by Orature... Should we be restricting that possibility here? I know there's other things currently preventing that possibility (source needs to be contained in the rc, for one), but is there anything specific that this string gives you that we need that check here (or is there a better way? Such as updating the table to contain a "derived project" column?)

PR #36, opened 2020-01-30
Jbs find recording bug

jsarabia:
Removes the volume bar for now (causes exceptions)

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/36)


aunger:
This fails SonarCloud. I guess the argument against committing commented-out code is that's what version control is for.

PR #38, opened 2020-02-03
Add Code Signing to the install4j build

danparisd:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/38)


danparisd:
We might want to push and tag and make this a release (with a patch version bump) so that there is a signed exe available on github.

PR #44, opened 2020-02-10
Jbs delete project

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/44)


jsarabia:
This uses the CSS from Thomas; He's setting variables in root.css. Sonarcloud is complaining but I'm not sure what this should be?

PR #52, opened 2020-02-27
Added Exception Dialog View

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/52)


mXaln:
___
*[jvm/workbookapp/src/main/resources/Messages_fr.properties, line 88 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/52#-M173oeBEvUjN3I-eOUO:-M1BAUbS5HQDqqDxO_Ql:b47s19m) ([raw file](https://github.com/bible-translation-tools/otter/blob/33c9969d5429c6273e61d479b34bea9a55b8263c/jvm/workbookapp/src/main/resources/Messages_fr.properties#L88)):*

this text is English, but the properties file is French. Please don't put English text here, we might forget that it needs to be translated because at a brief glance it looks like it already is.
</blockquote></details>

Should I leave those keys empty or not to add keys that don't have a translation?

mXaln:
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/skins/ExceptionDialogSkin.kt, line 63 at r3](https://reviewable.io/reviews/bible-translation-tools/otter/52#-M1asFq39zXsnUrrGVxV:-M1awR3T6DrfASyNnL35:b-bu91xm) ([raw file](https://github.com/bible-translation-tools/otter/blob/f87561ef7d03a2b7070c345844e61a21bbddaf62/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/skins/ExceptionDialogSkin.kt#L63)):*

> ```
>                     .then(
>                         FontIcon("gmi-expand-less").apply {
>                             styleClass.add("btn__icon")
>                         }
>                     )
> ```

you could also save both FontIcons to variables above and just use the variable in the then/otherwise. Currently this allocates a new object every time you switch between showing more and showing less, rather than reusing two of them.
</blockquote></details>

Updated

PR #56, opened 2020-03-18
Jira issue: OT119

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/56)

### DIFF
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt
@@ -25,9 +25,7 @@ class ResourceCardFragment(private val item: ResourceCardItem) : Fragment() {
 
     init {
         root.apply {
-            isFillHeight = false
             alignment = Pos.CENTER_LEFT
-            maxHeight = 50.0
 
             vbox {
                 spacing = 3.0
@@ -48,9 +46,8 @@ class ResourceCardFragment(private val item: ResourceCardItem) : Fragment() {
                     )
                 }
                 text(item.title) {
-                    wrappingWidth = 700.0
+                    wrappingWidthProperty().bind(navigator.root.widthProperty().divide(1.5))
                 }
-                maxWidth = 150.0
             }
 
             region {
@@ -63,7 +60,6 @@ class ResourceCardFragment(private val item: ResourceCardItem) : Fragment() {
                     secondaryColor = Color.WHITE
                     isHighlightedProperty.bind(isCurrentResourceProperty)
                     graphic = MaterialIconView(MaterialIcon.APPS, "25px")
-                    maxWidth = 500.0
                     text = messages["viewRecordings"]
                     action {
                         item.onSelect()
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt
@@ -47,7 +47,9 @@ class ResourceCardFragment(private val item: ResourceCardItem) : Fragment() {
                         }
                     )
                 }
-                text(item.title)
+                text(item.title) {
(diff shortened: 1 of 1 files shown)


mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt, line 51 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/56#-M2milurEFRF30joNoAb:-M2mzCII-KcaZ98NPfA8:bqupyih) ([raw file](https://github.com/bible-translation-tools/otter/blob/7dd0a8c8a8345b9fe3e8953f8b6fe367045703fd/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt#L51)):*

should this be instead based around some fraction of the width property of the container?
</blockquote></details>

the container's (vbox) max width is 150. If I bind wrappingWidth to vbox max witdh, it will be very narrow. Should I increase vbox's max width?

mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt, line 51 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/56#-M2milurEFRF30joNoAb:-M2n-1ew8hV4QF0m4uRF:bbdnh5) ([raw file](https://github.com/bible-translation-tools/otter/blob/7dd0a8c8a8345b9fe3e8953f8b6fe367045703fd/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt#L51)):*
<details><summary><i>Previously, mXaln (Maxim Murzamuratov) wrote…</i></summary><blockquote>

the container's (vbox) max width is 150. If I bind wrappingWidth to vbox max witdh, it will be very narrow. Should I increase vbox's max width?
</blockquote></details>

```
text(item.title) {
    wrappingWidthProperty().bind(this@vbox.maxWidthProperty())
}
maxWidth = 700.0
```

mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt, line 51 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/56#-M2milurEFRF30joNoAb:-M37Xxwf4DZPmFxDQDwY:b-ri6n1x) ([raw file](https://github.com/bible-translation-tools/otter/blob/7dd0a8c8a8345b9fe3e8953f8b6fe367045703fd/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt#L51)):*

I suspect both probably need to be fixed. I'm not exactly sure that it makes sense that the text wrap would ever be more than the width of the container itself? To me that sounds like if it's working and not cutting off text beyond the containers width, thus bleeding over into other containers, then it's kinda working by accident?
</blockquote></details>

Why do you think that wrapping width could be more than the max width of the container?

mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt, line 51 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/56#-M2milurEFRF30joNoAb:-M3A5QNy-kWAY_a9QYR-:be30trd) ([raw file](https://github.com/bible-translation-tools/otter/blob/7dd0a8c8a8345b9fe3e8953f8b6fe367045703fd/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/controls/resourcecard/view/ResourceCardFragment.kt#L51)):*

originally you have wrappingwidth = 700 and maxwidth = 150, which is what I was confused by.

I'm mostly okay with the code in your comment your comment, but why are you wrapping on the max width and not the actual width of the vbox?

I assume 700 is so it doesn't go beyond that and the cards are centered?
</blockquote></details>

vbox doesn't have a preferred width, so it stretches along with its children. It has the maximum length (700) and its child (text) will be wrapped to 700 if the text exceeds this value, thus will take all the maximum space of its parent. Hope it makes any sense. Why 700? I tried to run the app on resolution 1280x720 and this was the optimal choice. I'm not sure if it's possible to make wrappingWidth to fill entire blank space without specifying a certain value.

PR #59, opened 2020-03-20
Remember position of the last selected element and scroll to it on re…

mXaln:
…turn

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/59)

### DIFF
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/resources/view/ResourceListFragment.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/resources/view/ResourceListFragment.kt
@@ -22,7 +22,14 @@ class ResourceListFragment : Fragment() {
             }
         )
         add(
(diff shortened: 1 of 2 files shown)


mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/resources/viewmodel/ResourceListViewModel.kt, line 39 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/59#-M36ZcxH0clIF4-f5IlE:-M37WjDN5FAzr4lI31Uf:b-bw8e1h) ([raw file](https://github.com/bible-translation-tools/otter/blob/01e91440de8aabdb354f21b510d1652eabf46b00/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/resources/viewmodel/ResourceListViewModel.kt#L39)):*

One purpose of rx is to do things async so potential database transactions are not done on the UI thread. I know this might be cached and so a blockingFirst might not be a big deal, but if it does have to get resources from the database, this would block the UI thread.

It would be preferable to structure this in an async way
</blockquote></details>

Updated!

PR #70, opened 2020-04-06
Fix uninstall

danparisd:
This adds and/or fixes deletion of the user's Orature files on uninstall for Windows and Linux. The linux file deletion is a little hacky because dpkg has to be run as root/sudo and therefore there is no direct access to the current user's home directory. Also, we want to remove all users files on uninstall, so this does that. Open to other methods, this was just what seemed like it would work (and it did).

I also removed a non-working desktop shortcut script from the linux installer.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/70)

### DIFF
--- a/jvm/workbookapp/otter.install4j
+++ b/jvm/workbookapp/otter.install4j
@@ -307,7 +307,7 @@ return console.askYesNo(message, true);
               <action id="29" beanClass="com.install4j.runtime.beans.actions.UninstallFilesAction" actionElevationType="elevated" rollbackBarrierExitCode="0" />
               <action id="257" beanClass="com.install4j.runtime.beans.actions.files.DeleteFileAction" actionElevationType="elevated" rollbackBarrierExitCode="0">
                 <serializedBean>
-                  <property name="files" type="array" class="java.io.File" length="4">
+                  <property name="files" type="array" class="java.io.File" length="2">
                     <element index="0">
                       <object class="java.io.File">
                         <string>${installer:sys.appdataDir}${installer:sys.fileSeparator}Orature</string>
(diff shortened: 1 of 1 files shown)


jvm/workbookapp/otter.install4j
```diff
@@ -315,7 +315,7 @@ return console.askYesNo(message, true);
                     </element>
                     <element index="1">
                       <object class="java.io.File">
-                        <string>${installer:sys.userHome}/.config/Orature</string>
+                        <string>${installer:sys.userHome}${installer:sys.fileSeparator}Orature</string>
```
danparisd, line 318: This looks like I changed `/.config/` to `${installer:sys.fileSeparator}` but that's not what I did. I removed the entire `/.config/` line because it wasn't functional. That was for linux and is fixed with the new script at line 567. I then added the `<string>${installer:sys.userHome}${installer:sys.fileSeparator}Orature</string>` line for windows because there are files in two directories in windows.

jvm/workbookapp/otter.install4j
```diff
@@ -564,7 +564,7 @@ return console.askYesNo(message, true);
     </windows>
     <linuxDeb name="Linux Deb Archive" id="158">
       <jreBundle jreBundleSource="preCreated" includedJre="linux-amd64-11.0.4.tar.gz" manualJreEntry="true" />
-      <postInstallScript>cp /opt/${compiler:sys.shortName}/${compiler:sys.shortName}.desktop ~/Desktop/</postInstallScript>
+      <postUninstallScript>for i in $(ls -d -1 /home/**); do rm -rf $i/.config/Orature; done</postUninstallScript>
```
danparisd, line 567:
```suggestion
      <postUninstallScript>for i in $(ls -d -1 /home/**); do rm -rf $i/.config/Orature && rm -rf $i/Orature; done</postUninstallScript>
```

PR #82, opened 2020-04-17
New property for resource

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/82)

### DIFF
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeRecorder.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeRecorder.kt
@@ -19,7 +19,7 @@ class InitializeRecorder(
 ) : Installable {
 
     override val name = "RECORDER"
-    override val version = 1
+    override val version = 2
 
     val log = LoggerFactory.getLogger(InitializeRecorder::class.java)
 
@@ -54,7 +54,7 @@ class InitializeRecorder(
             AudioPluginData(
                 0,
                 "OtterRecorder",
-                "1.0.0",
+                "1.0.1",
                 canEdit = false,
                 canRecord = true,
                 executable = jar.absolutePath,
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeRecorder.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeRecorder.kt
@@ -19,7 +19,7 @@ class InitializeRecorder(
 ) : Installable {
 
     override val name = "RECORDER"
-    override val version = 2
+    override val version = 3
 
     val log = LoggerFactory.getLogger(InitializeRecorder::class.java)
 
@@ -54,7 +54,7 @@ class InitializeRecorder(
             AudioPluginData(
                 0,
                 "OtterRecorder",
-                "1.0.1",
+                "1.0.2",
                 canEdit = false,
                 canRecord = true,
                 executable = jar.absolutePath,
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeRecorder.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeRecorder.kt
@@ -19,7 +19,7 @@ class InitializeRecorder(
 ) : Installable {
 
     override val name = "RECORDER"
-    override val version = 3
+    override val version = 1
 
     val log = LoggerFactory.getLogger(InitializeRecorder::class.java)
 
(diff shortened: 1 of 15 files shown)


mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/takemanagement/viewmodel/AudioPluginViewModel.kt, line 98 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/82#-M57inSUCY_6D0jUWl5t:-M58409kEpZQEuWW9SWe:b3fkhg7) ([raw file](https://github.com/bible-translation-tools/otter/blob/3b42aa911814d6902e51e3ad8bd8f8e0c0bf0f6b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/takemanagement/viewmodel/AudioPluginViewModel.kt#L98)):*

I'm not happy with this code being in the app at all; but have accepted its existence. I will only allow it in one place though; given that it SHOULD be provided by the resource container. In the future I hope that will change, and thus only one place should have to be refactored. 

Since this when statement already exists somewhere, the data needs to be in something like the recordable object and passed in through a parameter. It doesn't belong in the AudioPluginViewModel
</blockquote></details>

Moved to WorkbookRepository

mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/takemanagement/viewmodel/AudioPluginViewModel.kt, line 98 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/82#-M57inSUCY_6D0jUWl5t:-M6aCtI45y6sQ-8ChJn_:bduh0ov) ([raw file](https://github.com/bible-translation-tools/otter/blob/3b42aa911814d6902e51e3ad8bd8f8e0c0bf0f6b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/takemanagement/viewmodel/AudioPluginViewModel.kt#L98)):*

I understand your thinking and you're on the right track. The repository is where data is inserted and pulled out; so when we get a workbook out of the repository it would pass through the code there and the domain model objects will have the proper values.

This suggests however that data inserted into the repository was incomplete, and we have hardcoded the data that needs to be inserted and do this every time the data is pulled out.

We admit we do have to hardcode the data, because it's a limitation of the Resource Container. We presumably will support other container formats (such as Scripture Burrito) and they're all ultimately mapped into our own schema that gets stored in the repository. So, since we have to take one container format, map it to our data model, and then insert it into the repository, the cleanest and best place to insert this is in the code that is specific to the Door43 Resource Container. That would be the code that imports specifically that kind of data format.

If we do that, we're ultimately sanitizing the data as it comes in, and then we know the data stored in the repository is sufficient for app use.

Similarly, if Scripture Burrito is one day supported, and if it has any sort of deficiencies that we have to compensate for, we would put that code in its import classes.

Also, if the specs of any of these kinds of formats change, their respective imports likely need to change as well. Maybe we need to add more data as a result, or maybe the updates remove the need for us to hardcode things like that. When we see a change to a spec, the first thing we should do is look to the import code for those formats and determine what we need to do to support the latest version. So this code would present itself in the exact place we'd be looking.
</blockquote></details>

Where can I find those import classes?

mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/takemanagement/viewmodel/AudioPluginViewModel.kt, line 41 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/82#-M57jREK7O5DWKt3-toe:-M6dxTUr1Yn2rUqxXWDM:b-bu91xm) ([raw file](https://github.com/bible-translation-tools/otter/blob/3b42aa911814d6902e51e3ad8bd8f8e0c0bf0f6b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/takemanagement/viewmodel/AudioPluginViewModel.kt#L41)):*

I feel like moving the variable out to be a nullable field of the class to avoid passing a parameter is an even more unclean solution that carries all the same issues, and introduces more potential null issues and state. Note there's a bug here. If you record a resource, then resource gets set in the viewmodel. If you then record something else such as a verse, the if check fails so resource is not updated, and when the parameters are created the old value is still used and you get a resource value that *should* be null.

It seems to me that there should be an activeResourceProperty in the workbookviewmodel like there is with chapter and chunks. That's at least consistent with the rest of them.

Alternatively, Chapter, Chunk, and Resource should instead all be parameters passed into record and we shouldn't be accessing any of them from the workbookviewmodel.
</blockquote></details>

Updated

PR #85, opened 2020-05-08
Select just created audio plugin as default

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/85)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/plugins/CreatePlugin.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/plugins/CreatePlugin.kt
@@ -1,18 +1,17 @@
 package org.wycliffeassociates.otter.common.domain.plugins
 
-import io.reactivex.Completable
+import io.reactivex.Maybe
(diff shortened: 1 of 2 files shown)


mXaln:
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/addplugin/viewmodel/AddPluginViewModel.kt, line 76 at r1](https://reviewable.io/reviews/bible-translation-tools/otter/85#-M7D-4hi7o_k8XZa3X2p:-M7DLjti3vOR9txAP8Hy:bi7qdz0) ([raw file](https://github.com/bible-translation-tools/otter/blob/36c4af4a4351ad5a7a9e32224a46c815cea01579/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/addplugin/viewmodel/AddPluginViewModel.kt#L76)):*

is this selecting this plugin regardless of what it was added for? (As in, if it can only record, it will be set as the editor too? Such as OratureRecorder?)
</blockquote></details>

The actual check whether it's a recording app or editing app or both happens down in the AudioPluginRepository in setRecorderData and setEditorData methods.

PR #100, opened 2020-06-15
Display source text

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/100)


common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/Chapter.kt
```diff
@@ -18,4 +19,22 @@ data class Chapter(
 
     override val contentType: ContentType = ContentType.META
     override val children: Observable<BookElement> = chunks.cast()
+
+    override val textItem
+        get() = textItem()
+
+    private fun textItem(): TextItem {
+        val initTextItem = TextItem("", MimeType.USFM)
```
jsarabia, line 27: What if it's not USFM?

PR #108, opened 2020-06-23
Removed redundant projects load from the init function

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/108)


mXaln:
This is an optimization fix. The projects are loaded twice initially. First when ProjectGridViewModel inititalized and also in ProjectGridFragment (onDock). https://github.com/Bible-Translation-Tools/otter/blob/e48f2a9967c040d2de6701732455a05253e66706/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/projectgrid/view/ProjectGridFragment.kt#L140

PR #130, opened 2020-07-16
Refactor source content

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/130)


mXaln:
Sorry, I couldn't split it into smaller pieces, otherwise it wouldn't build.

mXaln:
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/controllers/SourceContentController.kt, line 26 at r3](https://reviewable.io/reviews/bible-translation-tools/otter/130#-MCTEgB93sVpFjlVCn7W:-MCgi3Hc2MaeQTyhL6La:bgwqzv0) ([raw file](https://github.com/bible-translation-tools/otter/blob/028df202e4c717a36106a5a7beae2f9caf9e0472/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/controllers/SourceContentController.kt#L26)):*

this looks like it needs i18n templating?
</blockquote></details>

since the control doesn't have an access to i18n templating, I moved it out to WorkbookViewModel and SourceAudioFragment in recorderapp

PR #137, opened 2020-07-22
Add marker model

jsarabia:
Holds in progress changes to the marker state, computes next and previous marker from a given position

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/137)


jsarabia:
Requires PR#138

PR #140, opened 2020-07-27
Move audio in common

jsarabia:
Moves the audio/wav package out of IO.

This will be helpful for creating a jar of just this package that can be used as a library for other audio related apps.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/140)


mXaln:
@jsarabia As I can see there is only a package name change. So this PR could be approved. But my concern is that this PR revealed a lot of code smells, which I think should be fixed, like magic numbers (in some cases), empty default constructors. Let me know what you think.

PR #159, opened 2020-08-27
Fix numbering markers when placed

jsarabia:
Off by one on the index of the marker to insert.

Also adds adding one to markers after the marker being placed so that the order is correct if adding markers preceding placed markers.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/159)


mXaln (approved):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/159)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #169, opened 2020-09-04
Use one recording stream for volume and writing

jsarabia:
Use of multiple recorders was causing line exceptions due to race conditions.

This uses only one stream for both the volume bar and writing to a file, in which pausing sends a pause signal to the writer and "active" renderer, but the volume bar continues to get the recoding signal.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/169)


mXaln (approved):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/169)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #175, opened 2020-09-17
Remove onError for logging

jsarabia:
onError closes the stream and does catches the exception, which could cause unintended bugs, as onerror should be handled specific to each case.

Replaces them with onError for logging

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/175)


mXaln (approved):
> Reviewed 30 of 30 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/175)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #176, opened 2020-09-18
Return result when no plugin found

jsarabia:
Returns the result of no editor or recorder if a plugin is launched that doesn't exist which then triggers a snackbar to request adding a plugin

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/176)


mXaln (approved):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/176)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #209, opened 2020-12-04
Projects migration

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/209)

### DIFF
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeApp.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeApp.kt
@@ -67,7 +67,8 @@ class InitializeApp(
                         takeRepo,
                         languageRepo,
                         directoryProvider,
-                        zipEntryTreeBuilder
+                        zipEntryTreeBuilder,
+                        installedEntityRepo
                     )
                 )
 
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeApp.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeApp.kt
@@ -21,7 +21,8 @@ class InitializeApp(
     val collectionRepo: ICollectionRepository,
     val contentRepo: IContentRepository,
     val installedEntityRepo: IInstalledEntityRepository,
-    val zipEntryTreeBuilder: IZipEntryTreeBuilder
+    val zipEntryTreeBuilder: IZipEntryTreeBuilder,
+    val workbookRepository: IWorkbookRepository
 ) {
 
     private val logger = LoggerFactory.getLogger(InitializeApp::class.java)
@@ -68,7 +69,8 @@ class InitializeApp(
                         languageRepo,
                         directoryProvider,
                         zipEntryTreeBuilder,
-                        installedEntityRepo
+                        installedEntityRepo,
+                        workbookRepository
                     )
                 )
 
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt
@@ -5,9 +5,10 @@ import org.slf4j.LoggerFactory
 import org.wycliffeassociates.otter.common.domain.resourcecontainer.ImportResourceContainer
 import org.wycliffeassociates.otter.common.domain.resourcecontainer.project.IZipEntryTreeBuilder
 import org.wycliffeassociates.otter.common.persistence.IDirectoryProvider
-import org.wycliffeassociates.otter.common.persistence.config.Initializable
+import org.wycliffeassociates.otter.common.persistence.config.Installable
 import org.wycliffeassociates.otter.common.persistence.repositories.ICollectionRepository
 import org.wycliffeassociates.otter.common.persistence.repositories.IContentRepository
+import org.wycliffeassociates.otter.common.persistence.repositories.IInstalledEntityRepository
 import org.wycliffeassociates.otter.common.persistence.repositories.ILanguageRepository
 import org.wycliffeassociates.otter.common.persistence.repositories.IResourceContainerRepository
 import org.wycliffeassociates.otter.common.persistence.repositories.IResourceMetadataRepository
@@ -24,6 +25,7 @@ class InitializeProjects(
     private val languageRepo: ILanguageRepository,
     private val directoryProvider: IDirectoryProvider,
     private val zipEntryTreeBuilder: IZipEntryTreeBuilder,
+    private val installedEntityRepo: IInstalledEntityRepository,
     private val rcImporter: ImportResourceContainer = ImportResourceContainer(
         resourceMetadataRepo,
         resourceContainerRepo,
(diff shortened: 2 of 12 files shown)


jsarabia (changes requested):
> Reviewed 3 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/209)* status: 3 of 9 files reviewed, 3 unresolved discussions (waiting on @jsarabia and @mXaln)

___
*[assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt, line 193 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MNxZFa51WLjmwkqjGF-:-MNxZFa51WLjmwkqjGF0:bnwzud9) ([raw file](https://github.com/bible-translation-tools/otter/blob/0d4492cd96427c803152e814a03c7054891396e7/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt#L193)):*
> ```Kotlin
>     }
> 
>     private fun fetchProjects(): List<Workbook> {
> ```

The workbook as a data structure wasn't created until pretty far into the app's life. As such, we had already been loading projects from the database using collections.

I feel like we should have already refactored to actually make the WorkbookRepository the go to place for "projects" and getting projects should be functionality provided by it. Then we can use it in the future when we run into the situation of needing to load projects, and then writing code like this again.

For the most part, I assume that would just be moving this code to that class. I would request it be kept reactive though, so that the loading can happen async and when we use it we can make the decision whether or not to do it blocking. (we... probably should also be trying to make a better effort to keep things async though).
___
*[assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt, line 218 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MNx_a_O-mKtsUYbnLnf:-MNx_a_O-mKtsUYbnLng:bw7vlxw) ([raw file](https://github.com/bible-translation-tools/otter/blob/0d4492cd96427c803152e814a03c7054891396e7/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt#L218)):*
> ```Kotlin
>     }
> 
>     private fun createTempFile(name: String, extension: String): File {
> ```

is there a reason you're not just using "Files.createTempFile"?
___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/project/ProjectFilesAccessor.kt, line 92 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MNxQlzK4b9-dx6H17rA:-MNxQlzK4b9-dx6H17rB:b39532i) ([raw file](https://github.com/bible-translation-tools/otter/blob/0d4492cd96427c803152e814a03c7054891396e7/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/project/ProjectFilesAccessor.kt#L92)):*
> ```Kotlin
>         sources
>             .asSequence()
>             .map { it.path }
> ```

Can you explain what is going in in these first four lines and what the state of this chain is at the end of the map call?

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/209)* status: 3 of 12 files reviewed, 3 unresolved discussions (waiting on @jsarabia)

___
*[assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt, line 193 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MNxZFa51WLjmwkqjGF-:-MO12oh694SpDuZyuo7Z:b-896fix) ([raw file](https://github.com/bible-translation-tools/otter/blob/0d4492cd96427c803152e814a03c7054891396e7/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt#L193)):*

The workbook as a data structure wasn't created until pretty far into the app's life. As such, we had already been loading projects from the database using collections.

I feel like we should have already refactored to actually make the WorkbookRepository the go to place for "projects" and getting projects should be functionality provided by it. Then we can use it in the future when we run into the situation of needing to load projects, and then writing code like this again.

For the most part, I assume that would just be moving this code to that class. I would request it be kept reactive though, so that the loading can happen async and when we use it we can make the decision whether or not to do it blocking. (we... probably should also be trying to make a better effort to keep things async though).
</blockquote></details>

Done.
___
*[assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt, line 218 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MNx_a_O-mKtsUYbnLnf:-MO0y_7C-Ya60v4axe1k:benyifu) ([raw file](https://github.com/bible-translation-tools/otter/blob/0d4492cd96427c803152e814a03c7054891396e7/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeProjects.kt#L218)):*

is there a reason you're not just using "Files.createTempFile"?
</blockquote></details>

There is a reason. When I create a file using Files.createTempFile, the file name is something like en_tn879879.zip, and this file then  goes to rc->src dir. Then if user uninstalls the app and installs again, this file will be reimported from user home directory and a new name will like en_tn879879423423.zip (more random numbers) and so on, the name will grow. By creating temp directory instead I can avoid that.
___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/project/ProjectFilesAccessor.kt, line 92 at r2](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MNxQlzK4b9-dx6H17rA:-MO0zzvXD6AnQErh2q1H:b-kobmvi) ([raw file](https://github.com/bible-translation-tools/otter/blob/0d4492cd96427c803152e814a03c7054891396e7/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/project/ProjectFilesAccessor.kt#L92)):*

Can you explain what is going in in these first four lines and what the state of this chain is at the end of the map call?
</blockquote></details>

I create a mutableList with sourceMetadata item, then if linkedResource is not null, I add it to that list. Then, to use distinct function I convert the list to sequence.  I used mutableList, because for some reason I couldn't append the second item to the sequence. Unless I create a new sequence with the second item and concatinate that sequence with the first one. I got this code from ProjectImport class and left sequence stuff here. Probably I shouldn't. So removed toSequence method. Will do that with the list.

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/209)* status: 3 of 12 files reviewed, 2 unresolved discussions (waiting on @jsarabia and @mXaln)

___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt, line 81 at r3](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MO272CD3bdL8hGuYE12:-MO272CE5RkGDttOIIfE:b-arbac7) ([raw file](https://github.com/bible-translation-tools/otter/blob/c36b4f73a43737c8ccf54190f71ba70e69ae4b13/common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt#L81)):*
> ```Kotlin
>                 get(sourceProject, project)
>             }
>             .blockingGet()
> ```

Let's let this stay as an observable/single and leave it to the user of this method to determine whether or not they want to use blocking get

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/209)* status: 3 of 12 files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt, line 81 at r3](https://reviewable.io/reviews/bible-translation-tools/otter/209#-MO272CD3bdL8hGuYE12:-MO6gI624doynRq6X9Fe:b-896fix) ([raw file](https://github.com/bible-translation-tools/otter/blob/c36b4f73a43737c8ccf54190f71ba70e69ae4b13/common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt#L81)):*

Let's let this stay as an observable/single and leave it to the user of this method to determine whether or not they want to use blocking get
</blockquote></details>

Done.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/otter/209#-:-MO8ORcp4DjslxQ8ipCK:bnfp4nl)* status: 3 of 12 files reviewed, all discussions resolved

PR #277, opened 2021-04-30
Settings drawer created

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/277)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/277)* status: 0 of 17 files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/SettingsViewModel.kt, line 23 at r2](https://reviewable.io/reviews/bible-translation-tools/orature/277#-MZmPYzI8KaOxEnyLWWy:-MZmPYzI8KaOxEnyLWWz:b-frmt5s) ([raw file](https://github.com/bible-translation-tools/orature/blob/a74f899d4b7412ddbf6366ad3a5433bf1e84773a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/SettingsViewModel.kt#L23)):*
> ```Kotlin
> 
>     @Inject
>     lateinit var pluginRepository: IAudioPluginRepository
> ```

This is fine for now; but I think a lot of the work being done here belongs in a model object (one that should be injected like this is).

Viewmodels should probably not be interacting with repositories directly, so the methods here that do, should delegate to that model object (named... idk, PluginInteractor? not sure.)

Not a big deal, but it makes testing easier. The model object is what you want to test the most, and it allows for mocking the repository easily. Testing the logic if it sits in the viewmodel is more of an ordeal since it means more configuration with dependency injection- injecting the repository into the viewmodel, and it's harder to make the mock you want since it needs to go through dagger.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/277#-:-MZmqrtuALMFgYxcTr0m:bnfp4nl)* status: 0 of 17 files reviewed, 1 unresolved discussion (waiting on @mXaln)

PR #279, opened 2021-04-30
Log wav file name if exception raised from parsing metadata

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/279)


jsarabia:
Closing since I had to add in a hacky solution for log configuration. Need to think more about this before a PR.

PR #286, opened 2021-05-05
isNotEmpty to isNotBlank

jsarabia:
If the last line of the sql file is an empty line, the next block of code will try to execute a newline as a sql command, causing an exception to be thrown- this will only affect startup of a clean install; subsequent installs will avoid this function making the issue difficult to track down.

isNotBlank will filter out purely whitespace lines in addition to empty lists, fixing the issue.


mXaln (approved):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/286)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsarabia)

PR #298, opened 2021-05-26
Improve adding plugins for Mac

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/298)

PR #299, opened 2021-05-27
Improve adding plugins for Mac (#298)

jsarabia:
On MacOS, application binaries are deeper within their .app bundle in the /Applications or /System/Applications directories. These directories are treated as files in the JavaFX FileChooser, and as such cannot be used to select the binary.

We add the ability to edit the text in the path field, rather than rely solely on the result of the FileChooser.

Additionally, we further process the path before setting the value, detect if it is MacOS, and automatically navigate for the user.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/299)


mXaln (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/299)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsarabia)

PR #307, opened 2021-06-17
MP3 Playback Support

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/307)


mXaln (changes requested):
> Reviewed 11 of 25 files at r1, 19 of 20 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/307)* status: all files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt, line 28 at r3](https://reviewable.io/reviews/bible-translation-tools/orature/307#-Mcy2Ku53n7BJ7tSGnid:-Mcy2Ku53n7BJ7tSGnie:b-qigfn3) ([raw file](https://github.com/bible-translation-tools/orature/blob/a1d1453fa6d0aa70c91eca6668676fa02e033205/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt#L28)):*
> ```Kotlin
>         strategy = when (AudioFileFormat.of(file.extension)) {
>             AudioFileFormat.WAV -> WavFile(file, metadata as WavMetadata)
>             AudioFileFormat.MP3 -> MP3FileReader(file)
> ```

It's a bit confusing for me. Mp3FileReader implements both AudioFormatStrategy and AudioFileReader, where WavFileReader only implements AudioFileReader and WavFile implements AudioFormatStrategy. Maybe there should be also Mp3File class? To be the same as Wav thing?
___
*[common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/mp3/Mp3Metadata.kt, line 30 at r3](https://reviewable.io/reviews/bible-translation-tools/orature/307#-Mcy5VjP-hBM7hRM5Ul6:-Mcy5VjP-hBM7hRM5Ul7:b8wpl8w) ([raw file](https://github.com/bible-translation-tools/orature/blob/a1d1453fa6d0aa70c91eca6668676fa02e033205/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/mp3/Mp3Metadata.kt#L30)):*
> ```Kotlin
>                     val index = it.indices.find { it.number == 1 }
>                     index?.let {
>                         val position = (index.position.totalFrames / 75.0 * 44100.0).roundToInt()
> ```

What does 75.0 mean? Maybe have a variable or a constant for it? As well as for 44100.0

jsarabia (commented):
> Reviewed 5 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/307)* status: 27 of 36 files reviewed, 2 unresolved discussions (waiting on @jsarabia and @mXaln)

___
*[common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt, line 28 at r3](https://reviewable.io/reviews/bible-translation-tools/orature/307#-Mcy2Ku53n7BJ7tSGnid:-Mcz8SUx9lxNzQJXKRUP:bh09jmv) ([raw file](https://github.com/bible-translation-tools/orature/blob/a1d1453fa6d0aa70c91eca6668676fa02e033205/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt#L28)):*

It's a bit confusing for me. Mp3FileReader implements both AudioFormatStrategy and AudioFileReader, where WavFileReader only implements AudioFileReader and WavFile implements AudioFormatStrategy. Maybe there should be also Mp3File class? To be the same as Wav thing?
</blockquote></details>

The purpose of the strategy pattern is effectively to establish a generic public API solution who's functionality is delegated to specific implementations appropriate for a given input. 

So, we're defining an audio file, which provides some basic audio info, metadata, and can get a connection to a reader or writer regardless of which kind of audio file you provide as the input file.

So, the strategies are ultimately internal implementation details, and those details may not look the same as how another strategy looks due to the differences unique to those file formats. Maybe it will be better to make an MP3File separate from the file reader when writing is provided, but right now I don't know if there's a real reason to make the two look the same. I could probably also just move a lot of stuff from the WavFile class into the WavFileReader, since the WavFile was originally meant to be public API and now AudioFile is, but again, since it's internal implementation, I don't feel like it's urgent and also involves changing tests which are still currently valid.

I've changed the classes to specifically be internal though because the intention was to never allow someone to just make an MP3FileReader directly, the public API is to use: AudioFile(File("something.mp3")).reader()
___
*[common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/mp3/Mp3Metadata.kt, line 30 at r3](https://reviewable.io/reviews/bible-translation-tools/orature/307#-Mcy5VjP-hBM7hRM5Ul6:-MczB1R70EjdZ_Q8Uj2o:b-5i4inq) ([raw file](https://github.com/bible-translation-tools/orature/blob/a1d1453fa6d0aa70c91eca6668676fa02e033205/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/mp3/Mp3Metadata.kt#L30)):*

What does 75.0 mean? Maybe have a variable or a constant for it? As well as for 44100.0
</blockquote></details>

sure; it's the formula I took out of your cue manager library but that's fair. 75 has something to do with what a "frame" means in the context of cue files, which has something to do with CD sectors or something? Times in cues are "minute:second:frame" but frame is not the same as audio frames.

I've added some comments and source references.

mXaln (approved):
> Reviewed 11 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/307)* status: all files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt, line 28 at r3](https://reviewable.io/reviews/bible-translation-tools/orature/307#-Mcy2Ku53n7BJ7tSGnid:-MczJW_p0nfST02Tg5XB:b-pqov82) ([raw file](https://github.com/bible-translation-tools/orature/blob/a1d1453fa6d0aa70c91eca6668676fa02e033205/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt#L28)):*

The purpose of the strategy pattern is effectively to establish a generic public API solution who's functionality is delegated to specific implementations appropriate for a given input. 

So, we're defining an audio file, which provides some basic audio info, metadata, and can get a connection to a reader or writer regardless of which kind of audio file you provide as the input file.

So, the strategies are ultimately internal implementation details, and those details may not look the same as how another strategy looks due to the differences unique to those file formats. Maybe it will be better to make an MP3File separate from the file reader when writing is provided, but right now I don't know if there's a real reason to make the two look the same. I could probably also just move a lot of stuff from the WavFile class into the WavFileReader, since the WavFile was originally meant to be public API and now AudioFile is, but again, since it's internal implementation, I don't feel like it's urgent and also involves changing tests which are still currently valid.

I've changed the classes to specifically be internal though because the intention was to never allow someone to just make an MP3FileReader directly, the public API is to use: AudioFile(File("something.mp3")).reader()
</blockquote></details>

makes sense
___
*[common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/mp3/Mp3Metadata.kt, line 30 at r3](https://reviewable.io/reviews/bible-translation-tools/orature/307#-Mcy5VjP-hBM7hRM5Ul6:-MczJdAW0vEYWCu0dly_:b-q74txj) ([raw file](https://github.com/bible-translation-tools/orature/blob/a1d1453fa6d0aa70c91eca6668676fa02e033205/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/mp3/Mp3Metadata.kt#L30)):*

sure; it's the formula I took out of your cue manager library but that's fair. 75 has something to do with what a "frame" means in the context of cue files, which has something to do with CD sectors or something? Times in cues are "minute:second:frame" but frame is not the same as audio frames.

I've added some comments and source references.
</blockquote></details>

great!

PR #335, opened 2021-08-18
Add additional text, delete old localization files

jsarabia:
Adds the (hopefully) last text for Orature 1.0

Removes unofficial and massively outdated localization files so the official localization teams can start fresh.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/335)


AnonymousWalker (commented):
Can we remove (*.orature) from this? The file extension filter will include that when we config the ` chooseFile()  ` method

https://github.com/Bible-Translation-Tools/Orature/blob/3c5ec4e7f71a0c1467e60219826843f796b2a941/jvm/workbookapp/src/main/resources/Messages_en.properties#L245

AnonymousWalker (approved):

PR #337, opened 2021-08-18
Chunk Page refactoring

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337)


jsarabia (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337)* status: 0 of 29 files reviewed, 10 unresolved discussions (waiting on @mXaln)

___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/device/IAudioMerger.kt, line 1 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSjaSTDCWa8SvZ1r2X:-MiSjaSTDCWa8SvZ1r2Y:beahck) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/common/src/main/kotlin/org/wycliffeassociates/otter/common/device/IAudioMerger.kt#L1)):*
> ```Kotlin
> package org.wycliffeassociates.otter.common.device
> ```

Firstly, new files need a license as a header as per the GPL: https://www.gnu.org/licenses/gpl-howto.html (there is a gradle task called license format I think what does an okay job of trying to do this but sometimes it fails and we need to check manually before committing).

I assume you added this to the device package because it deals with audio stuff and most of the device package has to do with audio currently. I also assume you were trying to follow patterns of how these components work (defining an interface, implementing the interface, and instantiating via dependency injection).

The device layer is a very specific layer in the architecture for interacting with hardware devices. So far, this includes the microphone and speakers. The problem is, hardware access works differently on Android from JVM desktop and javax.sound libraries cannot be used on Android. Likewise, the libraries in Android are not available on desktop. We would prefer to not have to deal with the details of this in Orature, Orature just wants to be able to say "hey, I need an audio recorder to start recording." This particular package defines an interface for Orature to use some generic audio recorder, and the implementation provides the actual mechanism to do that in a platform specific way. The reason we need Dependency Injection is that it allows us a consistent way to get access to those objects. Some might take some involved and esoteric configuration to get started, and Orature just wants to say "inject/give me an audio recorder." It then allows for a specific and proper place to handle how to build the object, how to scope the object, and, in the case of platforms, which implementation to provide. 

The domain layer can be considered synonymous with "Orature code"- it's the code that completes tasks that people use Orature to accomplish. The UI is merely a rendering of the state of those tasks and providing a way to input data to perform the tasks, as well as a way to trigger execution of those tasks). So, users want to (in this case) combine a bunch of take files to get a chapter recording (this is a "use case" of the software, the task portion of it is domain code and should address the specific requests of a user story. The UI doesn't do the work, but merely gives the user a way to input the right data into this use case, execute it, and see the result.)

The Domain might take in some interfaces in its constructor of things it uses to accomplish tasks, and this allows us to write our use cases in the common package (which this already can do, since it uses no platform dependencies). But this reinforces the desire for a common package and how architecture can help us be more cross platform- as the domain code should be the bulk of what makes Orature do the stuff that users use Orature to do, the majority of the app itself becomes cross platform by nature- then, porting it to other platforms merely involves implementing interfaces of devices, file system locations, how to persist data (the implementation could move from using jooq to MongoDB if we desired, for example, and the rest of the app would show no perceived difference, and none of the rest of the code would change since the app just talks to the repositories), and, finally, the UI (which should merely be displaying data and receiving input to provide to domain objects- and even the UI might be able to be common soon with jetpack compose anyway).

We don't need to implement this in different ways, and we don't need to inject it. The ViewModel of the View this task is performed on should instantiate it and use it. 
___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/device/IAudioMerger.kt, line 7 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSxAXODFyYh1mobEV3:-MiSxAXODFyYh1mobEV4:b-1m4feh) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/common/src/main/kotlin/org/wycliffeassociates/otter/common/device/IAudioMerger.kt#L7)):*
> ```Kotlin
> 
> interface IAudioMerger {
>     fun merge(files: List<File>): Maybe<File>
> ```

I understand making this an observable because it allows for easily running this in the background, showing a progress spinner while the task runs, and easily closing the spinner when the task is done given that this might take some time.

That said, I don't think that a Maybe is the right approach. I don't know why files would be empty (and maybe that's more suitable as an exception if it is an error?) given that the compile option is only available when users have selected a take for each verse in a chapter, but even if there are empty files, should that matter? Even if they're all empty, could we not just produce an empty file of the appropriate file type? An empty wav file for example? which could then be opened in Ocenaudio like we already do, in which the user could add their data there?
___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt, line 383 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSkRuP8NhDfeVtEkVp:-MiSkRuP8NhDfeVtEkVq:b-vt47yu) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt#L383)):*
> ```Kotlin
>                 do {
>                     content.selectedTake = it.value?.let { wbTake -> takeMap[wbTake] }
>                 } while(content.selectedTake?.id == 0)
> ```

We should never use a busy wait as a solution to race conditions; it uses a lot of resources, could result in deadlocks, could result in significant lag etc. What is the problem you're encountering? Could synchronization solve it?
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/controllers/AudioPlayerController.kt, line 65 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSGAK2-LVCqpYnpOrX:-MiSGAK3A-OIHSZCHb3Z:b2ykkyi) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/controllers/AudioPlayerController.kt#L65)):*
> ```Kotlin
>     }
> 
>     fun load(player: IAudioPlayer) {
> ```

these changes (the whole file) should come in a separate bug fix PR; since this is all unrelated to the new UI

(and I have changed a lot of this stuff substantially in some of the code I'm working on, so I need to work through a massive merge conflict with your audio code in the other PR to begin with)
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 19 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSG6L7AmeFRR5_fzgb:-MiSG6L7AmeFRR5_fzgc:bd2kjl6) ([raw file](https://github.com/bible-translation-tools/orature/blob/6b93f7e45b5e5bad317ed0f1f0b68536eaf5b7f6/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L19)):*
> ```Kotlin
>  * along with Orature.  If not, see <https://www.gnu.org/licenses/>.
>  */
> package org.wycliffeassociates.otter.jvm.controls.media
> ```

Please don't remove the license from files
___
*[jvm/device/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioMerger.kt, line 8 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSHfi00eoD0W1rD_a1:-MiSHfi00eoD0W1rD_a2:b5zr157) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/device/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioMerger.kt#L8)):*
> ```Kotlin
> import java.io.File
> 
> class AudioMerger : IAudioMerger {
> ```

This would be better named as "ConcatenateAudio"
___
*[jvm/device/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioMerger.kt, line 14 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSI7sb5V12UWsekuEh:-MiSI7sc05x_ItN7QpBz:blbm0cr) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/device/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioMerger.kt#L14)):*
> ```Kotlin
>             if (files.isNotEmpty()) {
>                 val inputFile = AudioFile(files.first())
>                 val tempFile = File.createTempFile(
> ```

I would prefer that we move to using temporary files provided by the DirectoryProvider based temp directory; exceptions can result in a "deleteOnExit" call not being executed, leaving a large audio file in the users temp directory, and our users may not be able to track down temp directories for clearing up storage. 

Our temp directory we know about, and can clean up on application start every time.
___
*[jvm/device/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioMerger.kt, line 27 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSJ5bp20tRoJUOQizR:-MiSJ5bp20tRoJUOQizS:bdemoec) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/device/src/main/kotlin/org/wycliffeassociates/otter/jvm/device/audio/AudioMerger.kt#L27)):*
> ```Kotlin
>                 outputFile.writer(append = true).use { os ->
>                     files.forEach { file ->
>                         os.write(file.readBytes())
> ```

This is erroneous.

AudioFiles are binary files adhering to various format specifications. A wav file, for example, is a PCM stream in a RIFF container; it begins with a header providing information on file size, audio configuration (such as samplerate, bitrate, channels), size of the just audio section, etc. So, it begins with a header, and potentially ends with metadata. MP3 files are a bunch of concatenated frames containing about 1000 audio samples and each frame begins with a header. 

The AudioFile writer expects that the bytes being written are a PCM stream, and handles writing a valid file of its format (as in, it handles writing metadata). 

This means that in the case of Wav files, you will be writing out headers and metadata blocks as audio samples. You'll probably get a click or pop between files as a result.

In the case of MP3 files, a naive concatenation would probably work IF you were using a raw output stream, based on MP3 files being the concatenation of a bunch of small segments, BUT because you are using the writer from the audio file, it will encode the already encoded data as PCM data, and mostly just result in a jumbled mess.

The appropriate solution here would be to load each file as an AudioFile, read data over the reader (which will decode it to PCM) and output to the AudioFile's writer, which will encode the PCM to the appropriate format.
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/HomePage.kt, line 36 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiT-Fhh8j_TM_rCtZVx:-MiT-Fhh8j_TM_rCtZVy:b-tvclrw) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/HomePage.kt#L36)):*
> ```Kotlin
> import tornadofx.*
> 
> class HomePage : View() {
> ```

Why are a lot of Fragments being changed to Views?

Will this have some unintended consequences? Will we get dirty viewmodels that were previously always new?
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt, line 171 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiT-iK83yQ3K3JhXWcQ:-MiT-iK83yQ3K3JhXWcR:btts7jz) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt#L171)):*
> ```Kotlin
>     }
> 
>     fun setWorkChunk() {
> ```

What is a "work chunk"?

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337)* status: 0 of 29 files reviewed, 10 unresolved discussions (waiting on @jsarabia and @mXaln)

___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/HomePage.kt, line 36 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiT-Fhh8j_TM_rCtZVx:-MiWQhLD7sfCPS1Tel3O:b8zxvqg) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/HomePage.kt#L36)):*

Why are a lot of Fragments being changed to Views?

Will this have some unintended consequences? Will we get dirty viewmodels that were previously always new?
</blockquote></details>

On the contrary, viewmodels are created only once and then the instance is reused, though a new fragment is created every time you dock it. And previously created fragments do not get garbage collected because they still have bound values that still are listening for viewmodel changes, so kind of a memory leak. Unlike views that are created only once and stay in the view stack, so whenever we dock them, they are reused. Hope that makes sense.
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt, line 171 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiT-iK83yQ3K3JhXWcQ:-MiWPvHs3ayBonevKYdC:b9xo81d) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt#L171)):*

What is a "work chunk"?
</blockquote></details>

WorkChunk is the chunk to work with when you click "Begin Translation" or "Continue Translation"

jsarabia (changes requested):
> Reviewed 7 of 29 files at r1, 5 of 14 files at r2, 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337)* status: 15 of 33 files reviewed, 5 unresolved discussions (waiting on @mXaln)

___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt, line 26 at r4](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiaU7ay-W9JETecKzWC:-MiaU7ay-W9JETecKzWD:b6hjnmx) ([raw file](https://github.com/bible-translation-tools/orature/blob/0153f5f7251f82b5cbe8a9fdd8c027589161b65c/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt#L26)):*
> ```Kotlin
> import java.io.File
> 
> class ConcatenateAudio(private val directoryProvider: IDirectoryProvider) {
> ```

Domain classes in general should be good candidates for easy testing since they're free of dependencies from the UI and frameworks and the parameters should be able to be mocked relatively easily.

Can you add a unit test for this function? Mp3 files are too unpredictable by nature, but it should be pretty easy to create a couple wav files (in the set up function, not committed) where you write a sequence of numbers (such as 1,2 in the first file, 3,4 in the second, 5,6 in the third) and then at the end of the merge, you should be able to read all of the data out of the file with the AudioFile.reader() and it should give you 1,2,3,4,5,6 (nothing more or less) and you should be able to verify the WavFile audioLength and dataLength fields as well.
___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt, line 41 at r4](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiaPWOC-zLjNXGmw8Do:-MiaPWOC-zLjNXGmw8Dp:b-p1laxc) ([raw file](https://github.com/bible-translation-tools/orature/blob/0153f5f7251f82b5cbe8a9fdd8c027589161b65c/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt#L41)):*
> ```Kotlin
>                 files.forEach { file ->
>                     val audioFile = AudioFile(file)
>                     val buffer = ByteArray(1024)
> ```

You're going to want a larger buffer; probably add a 0 at the end. Otherwise it's going to be too little data to get a good result with how compression works in mp3.
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt, line 171 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiT-iK83yQ3K3JhXWcQ:-MiaT_Bp0MeDbWR5iUxT:b565jm0) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt#L171)):*

WorkChunk is the chunk to work with when you click "Begin Translation" or "Continue Translation"
</blockquote></details>

I see. I'm not sure a great name for it, could you add a javadoc comment to the method to explain the concept and what you're doing here?

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337)* status: 12 of 36 files reviewed, 5 unresolved discussions (waiting on @jsarabia and @mXaln)

___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt, line 41 at r4](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiaPWOC-zLjNXGmw8Do:-MibLnBoCxE4dfvnR458:b-896fix) ([raw file](https://github.com/bible-translation-tools/orature/blob/0153f5f7251f82b5cbe8a9fdd8c027589161b65c/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt#L41)):*

You're going to want a larger buffer; probably add a 0 at the end. Otherwise it's going to be too little data to get a good result with how compression works in mp3.
</blockquote></details>

Done.
___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt, line 383 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSkRuP8NhDfeVtEkVp:-MibL_wk52OEuYHRQJDX:b-ip5z9n) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/common/src/main/kotlin/org/wycliffeassociates/otter/common/persistence/repositories/WorkbookRepository.kt#L383)):*

We should never use a busy wait as a solution to race conditions; it uses a lot of resources, could result in deadlocks, could result in significant lag etc. What is the problem you're encountering? Could synchronization solve it?
</blockquote></details>

I removed that since we're gonna autoselect new takes
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/controllers/AudioPlayerController.kt, line 65 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSGAK2-LVCqpYnpOrX:-MibLqgo6csiCtXSbBmc:b-896fix) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/controllers/AudioPlayerController.kt#L65)):*

these changes (the whole file) should come in a separate bug fix PR; since this is all unrelated to the new UI

(and I have changed a lot of this stuff substantially in some of the code I'm working on, so I need to work through a massive merge conflict with your audio code in the other PR to begin with)
</blockquote></details>

Done.
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 19 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiSG6L7AmeFRR5_fzgb:-MibLscZ-xFRAFVe4NRe:b-tkncc6) ([raw file](https://github.com/bible-translation-tools/orature/blob/6b93f7e45b5e5bad317ed0f1f0b68536eaf5b7f6/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L19)):*

Please don't remove the license from files
</blockquote></details>

Done and moved to a separate PR
___
*[jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt, line 171 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiT-iK83yQ3K3JhXWcQ:-MibMa4C4h83sKQFoMH7:b-896fix) ([raw file](https://github.com/bible-translation-tools/orature/blob/dd5befc182df3648830b213ca2c2e18bb39156c5/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChapterPageViewModel.kt#L171)):*

I see. I'm not sure a great name for it, could you add a javadoc comment to the method to explain the concept and what you're doing here?
</blockquote></details>

Done.

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337)* status: 12 of 37 files reviewed, 5 unresolved discussions (waiting on @jsarabia)

___
*[common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt, line 26 at r4](https://reviewable.io/reviews/bible-translation-tools/orature/337#-MiaU7ay-W9JETecKzWC:-MigOAji7nhKfXN3epSP:b-mxmm67) ([raw file](https://github.com/bible-translation-tools/orature/blob/0153f5f7251f82b5cbe8a9fdd8c027589161b65c/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/content/ConcatenateAudio.kt#L26)):*

Domain classes in general should be good candidates for easy testing since they're free of dependencies from the UI and frameworks and the parameters should be able to be mocked relatively easily.

Can you add a unit test for this function? Mp3 files are too unpredictable by nature, but it should be pretty easy to create a couple wav files (in the set up function, not committed) where you write a sequence of numbers (such as 1,2 in the first file, 3,4 in the second, 5,6 in the third) and then at the end of the merge, you should be able to read all of the data out of the file with the AudioFile.reader() and it should give you 1,2,3,4,5,6 (nothing more or less) and you should be able to verify the WavFile audioLength and dataLength fields as well.
</blockquote></details>

Done. But I'm not sure how to create MP3 files to test concatenation. I think it's not necessary, because there is no Writer for mp3 files in AudioFormatStrategy.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/337#-:-MigqkZJ1x9oK7aJ482Z:bnfp4nl)* status: 12 of 37 files reviewed, 1 unresolved discussion (waiting on @jsarabia)

PR #420, opened 2021-11-11
Fix memory leak when using verse marker app

AnonymousWalker:
The waveform images (minimap and center waveform) were not cleaned when leaving the marker app => memory leak
This PR resolves the memory leak by:
- Removing main waveform image nodes from the UI
- Replacing the use of ReplaySubject with PublishSubject, thus only begins to build and emit images when the UI is ready to subscribe.
- Removing listeners attached to minimap image property.
- Discard minimap image by setting ImageView's inner image to null

Additionally, this handles abrupt exit from the marker app (before the waveform finished rendering) with doAfterTerminate() instead of doFinally(), which avoids logging "Closed stream" error when dispose() is invoked during the running loop.

This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/420)


jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/420)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker and @jsarabia)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/waveform/WaveformImageBuilder.kt, line 77 at r2](https://reviewable.io/reviews/bible-translation-tools/orature/420#-MoYhWRa5N1As7Xhyat1:-MoYhWRa5N1As7Xhyat2:bg43i6c) ([raw file](https://github.com/bible-translation-tools/orature/blob/68521067870b90279d1c1d7e6e84ea6a2365da75/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/waveform/WaveformImageBuilder.kt#L77)):*
> ```Kotlin
>         width: Int = Screen.getMainScreen().platformWidth,
>         height: Int = Screen.getMainScreen().platformHeight
>     ): Subject<Image> {
> ```

Why does the subject need to be exposed as a public API?

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/420)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jsarabia)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/waveform/WaveformImageBuilder.kt, line 77 at r2](https://reviewable.io/reviews/bible-translation-tools/orature/420#-MoYhWRa5N1As7Xhyat1:-MoYk3_M7Uo6QfjP1XKO:b-gifs9h) ([raw file](https://github.com/bible-translation-tools/orature/blob/68521067870b90279d1c1d7e6e84ea6a2365da75/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/waveform/WaveformImageBuilder.kt#L77)):*

Why does the subject need to be exposed as a public API?
</blockquote></details>

yeah it should be reverted as I was debugging the code with that

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/420#-:-MosYBsp93v6s84mvkYr:bnfp4nl)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @mXaln)

PR #421, opened 2021-11-12
Jbs time stretch

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/421)


jvm/controls/src/main/resources/css/filtered-search-bar.css
```diff
@@ -35,43 +35,7 @@
     -fx-fill: -wa-text-input-fill;
 }
 
-.filtered-search-bar__menu {
-    -fx-pref-width: 88px;
-    -fx-pref-height: 48px;
-    -fx-padding: 0 10px;
-    -fx-background-color: -wa-filter-button-background;
```
AnonymousWalker, line 42: looks like removing these bugged Create Project filter menu (dark mode)

AnonymousWalker: image

jvm/controls/src/main/resources/css/control.css
```diff
@@ -530,3 +530,72 @@
...
+.wa-menu-button__list-item:hover .label {
+    -fx-text-fill: -wa-primary;
+}
+
+.wa-menu-button__list-item .label {
+    -fx-font-size: 20px;
+}
+
+.wa-menu-button__arrow .ikonli-font-icon {
+   -fx-icon-size: 12 !important;
```
AnonymousWalker, line 599: I would not recommend using !important since there is usually an alternative way of doing it. Using !important often implies that the styles selector is not implemented very well somewhere on top of this.

AnonymousWalker: and this wouldn't block the PR, we can refactor it later

AnonymousWalker (approved):

PR #428, opened 2021-11-24
Playback rate button update

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/428)

### DIFF
--- a/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/button/WaMenuButton.kt
+++ b/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/button/WaMenuButton.kt
@@ -1,0 +1,68 @@
+package org.wycliffeassociates.otter.jvm.controls.button
+
+import javafx.beans.property.SimpleObjectProperty
+import javafx.beans.property.SimpleStringProperty
+import javafx.collections.FXCollections
+import javafx.collections.ObservableList
+import javafx.geometry.Side
+import javafx.scene.control.Button
+import javafx.scene.control.ContextMenu
+import javafx.scene.control.MenuItem
+import javafx.scene.layout.HBox
+import org.kordamp.ikonli.javafx.FontIcon
+import org.kordamp.ikonli.materialdesign.MaterialDesign
+import tornadofx.*
+
+class WaMenuButton : Button() {
+    val buttonTextProperty = SimpleStringProperty()
+    val sideProperty = SimpleObjectProperty<Side>()
+    val items: ObservableList<MenuItem> = FXCollections.observableArrayList()
+
+    private var menu: ContextMenu? = null
+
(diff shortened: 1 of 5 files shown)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/428)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @jsarabia and @mXaln)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 317 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/428#-MpgTbCIBIZH7dJ2pJMf:-MpgTbCIBIZH7dJ2pJMg:b7zt9u0) ([raw file](https://github.com/bible-translation-tools/orature/blob/cde4cbb54cca09255a0507890e252bbfc874acab/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L317)):*
> ```Kotlin
>         return CustomMenuItem().apply {
>             val formattedValue = String.format("%.2fx", speed)
>             val title = when (isCustom) {
> ```

You can assign values to an if/else, you don't have to use when for booleans.

If you do, it's probably better to explicitly exhaust the when, and have true/false cases rather than true/else.

But at a certain point, for booleans, I'd recommend just using if/else

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/428)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @jsarabia)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 317 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/428#-MpgTbCIBIZH7dJ2pJMf:-Mpkpo57BX2Ao_o2eC2X:b-896fix) ([raw file](https://github.com/bible-translation-tools/orature/blob/cde4cbb54cca09255a0507890e252bbfc874acab/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L317)):*

You can assign values to an if/else, you don't have to use when for booleans.

If you do, it's probably better to explicitly exhaust the when, and have true/false cases rather than true/else.

But at a certain point, for booleans, I'd recommend just using if/else
</blockquote></details>

Done.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/428#-:-MpkxbkP8znDNTZC9HZW:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jsarabia)

PR #426, opened 2021-11-18
Playback rate UI

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/426)

PR #432, opened 2021-11-30
Fixed bug when tooltip wasn't showing for play buttons without text

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432)

### DIFF
--- a/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt
+++ b/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt
@@ -86,9 +86,7 @@ class SimpleAudioPlayer(
             addClass("btn", "btn--icon")
             textProperty().bind(playPauseTextBinding())
             tooltip {
-                textProperty().bind(audioPlayerController.isPlayingProperty.stringBinding{
-                    if (it == true) FX.messages["pause"] else FX.messages["play"]
-                })
+                textProperty().bind(playPauseTextBinding(true))
             }
             graphicProperty().bind(
                 audioPlayerController.isPlayingProperty.objectBinding { isPlaying ->
@@ -146,9 +144,7 @@ class SimpleAudioPlayer(
                 button.apply {
                     textProperty().bind(playPauseTextBinding())
                     tooltip {
-                        textProperty().bind(audioPlayerController.isPlayingProperty.stringBinding{
-                            if (it == true) FX.messages["pause"] else FX.messages["play"]
-                        })
+                        textProperty().bind(playPauseTextBinding(true))
                     }
                     graphicProperty().bind(
                         audioPlayerController.isPlayingProperty.objectBinding { isPlaying ->
@@ -169,14 +165,19 @@ class SimpleAudioPlayer(
         menuItems.setAll(createPlaybackRateMenu())
     }
 
-    private fun playPauseTextBinding(): StringBinding {
+    private fun playPauseTextBinding(hideButtonText: Boolean = false): StringBinding {
         return Bindings.createStringBinding(
             {
-                if (audioPlayerController.isPlayingProperty.value == true) {
-                    pauseTextProperty.value
-                } else {
-                    playTextProperty.value
+                val isPlaying = audioPlayerController.isPlayingProperty.value ?: false
+                var pauseText = pauseTextProperty.value
+                var playText = playTextProperty.value
+
+                if (hideButtonText) {
+                    pauseText = pauseText ?: FX.messages["pause"]
+                    playText = playText ?: FX.messages["play"]
                 }
+
+                if (isPlaying) pauseText else playText
             },
             audioPlayerController.isPlayingProperty,
             playTextProperty
--- a/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt
+++ b/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt
@@ -86,7 +86,7 @@ class SimpleAudioPlayer(
             addClass("btn", "btn--icon")
             textProperty().bind(playPauseTextBinding())
             tooltip {
-                textProperty().bind(playPauseTextBinding())
+                textProperty().bind(playPauseTextBinding(true))
             }
             graphicProperty().bind(
                 audioPlayerController.isPlayingProperty.objectBinding { isPlaying ->
@@ -144,7 +144,7 @@ class SimpleAudioPlayer(
                 button.apply {
                     textProperty().bind(playPauseTextBinding())
                     tooltip {
-                        textProperty().bind(playPauseTextBinding())
+                        textProperty().bind(playPauseTextBinding(true))
                     }
                     graphicProperty().bind(
                         audioPlayerController.isPlayingProperty.objectBinding { isPlaying ->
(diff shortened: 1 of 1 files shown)


jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt
```diff
@@ -165,12 +165,19 @@ class SimpleAudioPlayer(
         menuItems.setAll(createPlaybackRateMenu())
     }
 
-    private fun playPauseTextBinding(): StringBinding {
+    private fun playPauseTextBinding(tooltip: Boolean = false): StringBinding {
```
AnonymousWalker, line 168: do we need it as a parameter? I thought all interactive UI elements should have tooltip

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 168 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/432#-MplccSduXk-kpIvNZjY-r1-168:-Mpph_CqBhRqtNaOPHAi:brm4ljy) ([raw file](https://github.com/bible-translation-tools/orature/blob/dc1c0f2f8acb115139ae43fec034c283e867deee/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L168)):*
<details><summary><i>Previously, AnonymousWalker (Tony Anh Tran) wrote…</i></summary><blockquote>

do we need it as a parameter? I thought all interactive UI elements should have tooltip
</blockquote></details>

yes, but this parameter is applied only for play buttons, and some play buttons may not have text (just icon). So this playPauseTextBinding returns text based on button text or if it's just an icon button, it returns default text for tooltip only.

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker and @jsarabia)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 168 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/432#-MplccSduXk-kpIvNZjY-r1-168:-Mpv0MeWDK_0qaTf21hp:b-un2dv1) ([raw file](https://github.com/bible-translation-tools/orature/blob/dc1c0f2f8acb115139ae43fec034c283e867deee/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L168)):*

yes, but this parameter is applied only for play buttons, and some play buttons may not have text (just icon). So this playPauseTextBinding returns text based on button text or if it's just an icon button, it returns default text for tooltip only.
</blockquote></details>

I think the parameter name is confusing, because if I understand correctly; you're always going to have a tooltip, this is only guarding a fallback to default if the parameter is set (which is intended for use when the button is icon only).

If that's true, is there a reason why we'd want to guard it in an if statement anyway? Is there any harm in just always having the fallback if the text is null since we always want tooltips?

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker and @jsarabia)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 168 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/432#-MplccSduXk-kpIvNZjY-r1-168:-MpvGi0a-Cfpp-h7Gcps:b-nfm8p5) ([raw file](https://github.com/bible-translation-tools/orature/blob/dc1c0f2f8acb115139ae43fec034c283e867deee/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L168)):*

I think the parameter name is confusing, because if I understand correctly; you're always going to have a tooltip, this is only guarding a fallback to default if the parameter is set (which is intended for use when the button is icon only).

If that's true, is there a reason why we'd want to guard it in an if statement anyway? Is there any harm in just always having the fallback if the text is null since we always want tooltips?
</blockquote></details>

if we have a fallback when text is null, it means that every button will have the text. But we don't want the text for icon buttons, only tooltips. I could have created a separate binding for tooltips text, but it would be just duplicated code.

AnonymousWalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 168 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/432#-MplccSduXk-kpIvNZjY-r1-168:-MpvKvYX5DE282pVCZfF:b-4mc78f) ([raw file](https://github.com/bible-translation-tools/orature/blob/dc1c0f2f8acb115139ae43fec034c283e867deee/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L168)):*

if we have a fallback when text is null, it means that every button will have the text. But we don't want the text for icon buttons, only tooltips. I could have created a separate binding for tooltips text, but it would be just duplicated code.
</blockquote></details>

I believe when I applied the tooltip to every button, it didn't need any binding/relationship with the button text. Looks like separating the button text and tooltip is better than clumping them together and having condition when to show what

``` 
            tooltip {
                textProperty().bind(audioPlayerController.isPlayingProperty.stringBinding{
                    if (it == true) messages["pause"] else messages["play"]
                })
            }
```

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker and @jsarabia)

___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 168 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/432#-MplccSduXk-kpIvNZjY-r1-168:-MpvMMbeDnAFL6kxAhkY:b9vpdj5) ([raw file](https://github.com/bible-translation-tools/orature/blob/dc1c0f2f8acb115139ae43fec034c283e867deee/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L168)):*
<details><summary><i>Previously, AnonymousWalker (Tony Anh Tran) wrote…</i></summary><blockquote>

I believe when I applied the tooltip to every button, it didn't need any binding/relationship with the button text. Looks like separating the button text and tooltip is better than clumping them together and having condition when to show what

``` 
            tooltip {
                textProperty().bind(audioPlayerController.isPlayingProperty.stringBinding{
                    if (it == true) messages["pause"] else messages["play"]
                })
            }
```
</blockquote></details>

The text is not always pause and play, sometimes it's a custom text (Play Source, Pause Target Audio. etc..)

AnonymousWalker:
___
*[jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt, line 168 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/432#-MplccSduXk-kpIvNZjY-r1-168:-MpvOS_H-eODhH9ksXq6:b-e6rssi) ([raw file](https://github.com/bible-translation-tools/orature/blob/dc1c0f2f8acb115139ae43fec034c283e867deee/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/media/SimpleAudioPlayer.kt#L168)):*

The text is not always pause and play, sometimes it's a custom text (Play Source, Pause Target Audio. etc..)
</blockquote></details>

looks like the SimpleAudioPlayer is refactored to include text for the play button since #426. Initially, SAP only had an icon button for play/pause and I think that was what it should be. Is there a way we can apply inheritance/polymorphism for a SAP having text in the Play button?

https://github.com/Bible-Translation-Tools/Orature/commit/fdaa94d6bdfb230143f4278bef83ff8884b2f660#diff-b232a5213f3e4caebce88782455eacab8fa3e0f0b019e8ae41bf347ab1f35f4f

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/432#-:-MqKlt6e2TObl7ED3vzS:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

PR #480, opened 2022-01-12
Remove Deprecations

jsarabia:
Removes deprecated method calls.

toLowerCase is replaced with lowercase. This is chosen over lowercase(Locale.Default) due to the uses being slugs or file extensions, which, could be problematic if lowercase doesn't work as expected on different locales.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/480)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/utils/StringUtils.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/utils/StringUtils.kt
@@ -1,3 +1,21 @@
+/**
+ * Copyright (C) 2020-2022 Wycliffe Associates
+ *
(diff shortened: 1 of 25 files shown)


common/src/main/kotlin/org/wycliffeassociates/otter/common/utils/StringUtils.kt
```diff
@@ -0,0 +1,7 @@
+package org.wycliffeassociates.otter.common.utils
```
AnonymousWalker, line 19: I assume you would add the open source license here before merging

AnonymousWalker (approved):

AnonymousWalker (approved):

PR #485, opened 2022-01-18
Avoid duplicated css imports

AnonymousWalker:
importStylesheet() will just flood the stylesheet list without concern about duplicated entries. This PR adds a new wrapper function which prevents adding if a stylesheet has already imported.

This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/485)

PR #487, opened 2022-01-19
Chunking: Verbalize Step

jsarabia:
Adds the Verbalize step of chunking.

This is not connected to the main application, but part of a separate debug entrypoint for the Chunking App.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/487)

### DIFF
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/chunking/ChunkingDebugMain.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/chunking/ChunkingDebugMain.kt
@@ -20,6 +20,7 @@ package org.wycliffeassociates.otter.jvm.workbookapp.ui.screens.chunking
 
 import javafx.scene.layout.StackPane
 import javafx.stage.Stage
+import org.wycliffeassociates.otter.jvm.controls.styles.tryImportStylesheet
 import org.wycliffeassociates.otter.jvm.device.ConfigureAudioSystem
 import org.wycliffeassociates.otter.jvm.device.audio.AudioDeviceProvider
 import org.wycliffeassociates.otter.jvm.device.audio.DEFAULT_AUDIO_FORMAT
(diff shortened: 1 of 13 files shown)


jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/chunking/ChunkingDebugMain.kt
```diff
@@ -0,0 +1,64 @@
...
+
+fun main(args: Array<String>) {
+    launch<ChunkingDebugApp>(args)
+}
+
+class ChunkingDebugApp : App(ChunkingDebugView::class), IDependencyGraphProvider {
+    override val dependencyGraph = DaggerAppDependencyGraph.builder().build()
+
+    init {
+        importStylesheet(resources["/css/theme/light-theme.css"])
```
AnonymousWalker: Please use tryImportStylesheet() instead (see #485 for detail).

jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChunkingViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+                "Verbalize" -> {
+                    consumeStepColor.set(green)
+                    verbalizeStepColor.set(blue)
+                    chunkStepColor.set(gray)
+                }
+                "Chunking" -> {
+                    consumeStepColor.set(green)
+                    verbalizeStepColor.set(green)
+                    chunkStepColor.set(blue)
+                }
```
AnonymousWalker, line 60: I'd strongly recommend using css style classes for binding to the state of each step. This allows changing and especially theming easier.

An example I used was in RootView.bindThemeClassToRoot(), which basically toggles between different classes based on the property's value - onChange(). 

```
        viewModel.appColorMode.onChange {
            when (it) {
                ColorTheme.LIGHT -> {
                    root.addClass("light-theme")
                    root.removeClass("dark-theme")
                }
                ColorTheme.DARK -> {
                    root.addClass("dark-theme")
                    root.removeClass("light-theme")
                }
            }
        }
```
Then, we can just define what colors to be used with each style class in a particular theme css file.
In this case, maybe "-wa-consume-background", "-wa-verbalize-background" and "-wa-chunking-background"

AnonymousWalker (approved):

PR #509, opened 2022-03-02
fix release/prerelease bug

danparisd:
This should fix qa/test builds to push as pre-release. It looks like one of my commits didn't make it into the last PR, but I believe this should fix it. I also noticed that there would have been a bug if we had done a release build, so this should fix that too.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/509)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/509#-:-MxB9_VI3ULvDZKyqOxr:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #542, opened 2022-04-01
Make source content not focusable when text is small

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/542)


AnonymousWalker:
The upcoming changes in #544 will void this PR as we replace the source text with a list view of verses

mXaln:
Will break if merged with #544

AnonymousWalker:
Resolved by #551

PR #608, opened 2022-06-17
Fix reset chapter

AnonymousWalker:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/608)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/608)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker and @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/AssociatedAudio.kt` line 47 at r2](https://reviewable.io/reviews/bible-translation-tools/orature/608#-N5ameY164e4GOeKT07y:-N5ameY164e4GOeKT07z:b-nqljf7) ([raw file](https://github.com/bible-translation-tools/orature/blob/e358b7f920163693adfece86dcf88d689ac6bd62/common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/AssociatedAudio.kt#L47)):*
> ```Kotlin
>         Single.just(
>             getAllTakes()
>                 .filter { it.deletedTimestamp.value?.value == null }
> ```

I foresee this being a bug;

Let's say take 4 was scheduled to be deleted
A new take is created, take 3 is the new max after we filter out take 4, +1 means the new take is 4
take 4 file is overwritten
Close App
Open app, takes with the deleted timestamp set are deleted
new take 4 is now deleted

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/608)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/AssociatedAudio.kt` line 47 at r2](https://reviewable.io/reviews/bible-translation-tools/orature/608#-N5ameY164e4GOeKT07y:-N5eMFUg4etSXkP5NVq1:b-4cnils) ([raw file](https://github.com/bible-translation-tools/orature/blob/e358b7f920163693adfece86dcf88d689ac6bd62/common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/AssociatedAudio.kt#L47)):*

I foresee this being a bug;

Let's say take 4 was scheduled to be deleted
A new take is created, take 3 is the new max after we filter out take 4, +1 means the new take is 4
take 4 file is overwritten
Close App
Open app, takes with the deleted timestamp set are deleted
new take 4 is now deleted
</blockquote></details>

Thanks for the good catch

PR #611, opened 2022-06-27
Move code for markers into an interface

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/611)


jsarabia:
Moves shared code from similar ViewModels into an interface.

PR #633, opened 2022-07-25
Fix a memory leak related to plugins

jsarabia:
Each scope stores a map in tornadofx holding all Views and Fragments for that scope- as such, without deregistering, they will leak every time a plugin is opened.

Also updates tornadofx to a self hosted build which includes the latest changes including a fix to a Fragment related memory leak.

---
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/633)


AnonymousWalker (approved):

AnonymousWalker (approved):
Looks like it couldn't compile properly, could you check?

PR #656, opened 2022-08-16
Chunk storage tests

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/656)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/656)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jsarabia and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/chunking/ChunkAudioUseCaseTest.kt` line 38 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/656#-N9cCJ6c8lQ2mlTYTkv7:-N9cCJ6c8lQ2mlTYTkv8:b-32ibhh) ([raw file](https://github.com/bible-translation-tools/orature/blob/6e2ea9b126cb53a72edb0935d33316eeca28f165/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/chunking/ChunkAudioUseCaseTest.kt#L38)):*
> ```Kotlin
>         region = "Europe",
>         id = autoincrement
>     )
> ```

Languages, Resource Containers, Dublin Core...
A lot of this is just dumb data. These things take up a ton of lines of code in tests. They're in most tests.

Why do we need them in every test? Why can't they be mocked? Are they dependencies of things that should also just be mocked?

If they actually, genuinely, 100% need to be included, then why aren't they in one common test utils file and reused?

Tests are code, all the same style rules and things such as SOLID/DRY/etc should still apply.

jsarabia (commented):
The things to test here are:
source audio is copied to the RC location
cue file is copied to the RC location
source audio is updated with the new cue list

Which you're doing (good job!)

What I would recommend is noting that the workbook is only used because the use case is creating a projectfilesaccessor, and as such it would probably make more sense to refactor the constructor to take in that instead of the workbook. That would enable mocking the PFA directly, which would be more direct and involve less set up than mocking things to allow the code to create one.

A resource container is used as well, and it would probably be valuable to have a function to create an empty test RC in a utils method, and your mocked PFA could direct the use case to the test RC.

It might also be best to make the call to the function be done in setup, and then have three tests explicitly verifying the three things listed above.

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/656#-:-N9mS0zs4lh-3gHrpQVD:bz95olu)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jsarabia)

common/src/test/kotlin/org/wycliffeassociates/otter/common/TestUtils.kt
```diff
@@ -83,3 +99,150 @@ fun createTestWavFile(dir: File): File {
     return testFile
 }
 
+fun getDublinCore(resource: ResourceMetadata): DublinCore {
```
PurpleGuitar, line 102: I like these test utilities to generate test objects. However, I think the name is misleading -- "get" implies looking something up or retreiving it, whereas this function creates a new object.  I think "createDublinCore" or "createTestDublinCore" would be a better name for this function (and similarly for the other "get" functions).

common/src/test/kotlin/org/wycliffeassociates/otter/common/TestUtils.kt
```diff
@@ -83,3 +99,150 @@ fun createTestWavFile(dir: File): File {
...
+fun buildWorkbook(
+    db: WorkbookRepository.IDatabaseAccessors,
+    source: Collection,
+    target: Collection
+) = WorkbookRepository(
+    mock(),
+    db
+).get(source, target)
+
+fun createWavFile(dir: File, name: String, data: ByteArray): File {
```
PurpleGuitar, line 186: Is creating physical WAV files necessary for testing?  IO actions make tests slow and expensive.  Could this be mocked with FileStreams or something similar?

PurpleGuitar (approved):

PR #673, opened 2022-09-19
call sql.close() when not on macos

jsarabia:
M1 macs cannot execute this line, but are not bothered by deleting this file if it is open.

- - -
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/673)


AnonymousWalker (approved):
Thanks, this would keep Windows from holding the file opened

PR #691, opened 2022-10-14
Compile chapter content before export

AnonymousWalker:
When exporting a project as source audio, we first check and compile the chapters which are completed and ready.

This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/691)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt
@@ -43,6 +43,7 @@ import javax.inject.Inject
 abstract class ProjectExporter {
     @Inject
     lateinit var concatenateAudio: ConcatenateAudio
+
     @Inject
     lateinit var takeActions: TakeActions
 
@@ -98,13 +99,23 @@ abstract class ProjectExporter {
     ): Completable {
         return workbook.target
             .chapters
-            .filter {
-                // filter completed chapters which have not been compiled yet
-                it.audio.selected.value?.value == null &&
-                        it.chunks.all { chunk ->
-                            chunk.audio.selected.value?.value != null
-                        }.blockingGet()
-            }.flatMapCompletable { chapter ->
+            .filter { chapter ->
+                // filter chapter without selected take
+                chapter.audio.selected.value?.value == null
+            }
+            .flatMap { chapter ->
+                // filter chapter where all its content are ready to compile
+                chapter.chunks
+                    .all { chunk ->
+                        chunk.audio.selected.value?.value != null
+                    }
+                    .toObservable()
+                    .mapNotNull {
+                        if (it) chapter else null
+                    }
+            }
+            .flatMapCompletable { chapter ->
+                // compile the chapter
                 chapter.chunks
                     .mapNotNull { chunk -> chunk.audio.selected.value?.value?.file }
                     .toList()
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt
@@ -18,20 +18,35 @@
  */
 package org.wycliffeassociates.otter.common.domain.resourcecontainer.projectimportexport
 
+import io.reactivex.Completable
 import io.reactivex.Single
+import io.reactivex.schedulers.Schedulers
+import org.slf4j.LoggerFactory
 import org.wycliffeassociates.otter.common.data.primitives.Contributor
 import org.wycliffeassociates.otter.common.data.primitives.ResourceMetadata
+import org.wycliffeassociates.otter.common.data.workbook.Chapter
 import org.wycliffeassociates.otter.common.data.workbook.Workbook
+import org.wycliffeassociates.otter.common.domain.content.ConcatenateAudio
 import org.wycliffeassociates.otter.common.domain.content.FileNamer
+import org.wycliffeassociates.otter.common.domain.content.TakeActions
+import org.wycliffeassociates.otter.common.domain.content.WorkbookFileNamerBuilder
 import org.wycliffeassociates.otter.common.domain.resourcecontainer.project.ProjectFilesAccessor
+import org.wycliffeassociates.otter.common.utils.mapNotNull
 import org.wycliffeassociates.resourcecontainer.ResourceContainer
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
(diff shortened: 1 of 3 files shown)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/691)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt` line 71 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/691#-NEb1wTVEq2tMhqW1h8E:-NEb1wTVEq2tMhqW1h8F:b-gluzrj) ([raw file](https://github.com/bible-translation-tools/orature/blob/65dc58dd56eba1ec9ac9ec8ad63b909dc7493516/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt#L71)):*
> ```Kotlin
> 
>     /**
>      * Changes the file extension.
> ```

The method name is more informative than the comment. Obviously it changes the file extension, the method name says so. In fact, it's being "restored" implying the extension was removed or changed to something else.

Why was the extension removed?
Restore to what?
Why was it removed to begin with?
Why is this a protected method instead of private?
Who uses this method and why?

The purpose of comments needs to be to provide context and extra information to other developers that is not obvious. Paraphrasing the code typically does not provide value, unless the code is itself difficult to read (which, would ideally suggest a refactor to simplify the code rather than commenting, but sometimes complexity is unavoidable).

In the case of an abstract class, this method could be intended to be subclassed, and if that was the case, what is the context in which this method is called in order for a developer to subclass it properly?

Think about who is going to be reading comments, why, and what would be helpful for them; but also think about yourself and the kind of comment that you would hope you would find if you didn't write the code but needed to use or maintain it.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt` line 106 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/691#-NEb3NEO6nZXKiIThd1s:-NEb3NEO6nZXKiIThd1t:b-zge9f) ([raw file](https://github.com/bible-translation-tools/orature/blob/65dc58dd56eba1ec9ac9ec8ad63b909dc7493516/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt#L106)):*
> ```Kotlin
>             .filter {
>                 // filter completed chapters which have not been compiled yet
>                 it.audio.selected.value?.value == null &&
> ```

this expression is too long, please save the values to variables
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt` line 109 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/691#-NEb3Q_B8bh16GdCHB9v:-NEb3Q_B8bh16GdCHB9w:b-ove6m4) ([raw file](https://github.com/bible-translation-tools/orature/blob/65dc58dd56eba1ec9ac9ec8ad63b909dc7493516/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt#L109)):*
> ```Kotlin
>                         it.chunks.all { chunk ->
>                             chunk.audio.selected.value?.value != null
>                         }.blockingGet()
> ```

is it possible to transform this rx stream to flow naturally rather than have nested blocking calls?

AnonymousWalker:
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt` line 71 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/691#-NEb1wTVEq2tMhqW1h8E:-NEb4RHK8ARUbADTnuuB:b2ufvaz) ([raw file](https://github.com/bible-translation-tools/orature/blob/65dc58dd56eba1ec9ac9ec8ad63b909dc7493516/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt#L71)):*

The method name is more informative than the comment. Obviously it changes the file extension, the method name says so. In fact, it's being "restored" implying the extension was removed or changed to something else.

Why was the extension removed?
Restore to what?
Why was it removed to begin with?
Why is this a protected method instead of private?
Who uses this method and why?

The purpose of comments needs to be to provide context and extra information to other developers that is not obvious. Paraphrasing the code typically does not provide value, unless the code is itself difficult to read (which, would ideally suggest a refactor to simplify the code rather than commenting, but sometimes complexity is unavoidable).

In the case of an abstract class, this method could be intended to be subclassed, and if that was the case, what is the context in which this method is called in order for a developer to subclass it properly?

Think about who is going to be reading comments, why, and what would be helpful for them; but also think about yourself and the kind of comment that you would hope you would find if you didn't write the code but needed to use or maintain it.
</blockquote></details>

thanks for the feedback. My initial thought was to simplify the purpose of this function. The actual use of it is to change from `.zip ` to `.orature`

I can remove it then, since I couldn't think of a better way to comment here.

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/691)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt` line 106 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/691#-NEb3NEO6nZXKiIThd1s:-NEb6YsH9AXJ_qdu-kn7:b-896fix) ([raw file](https://github.com/bible-translation-tools/orature/blob/65dc58dd56eba1ec9ac9ec8ad63b909dc7493516/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt#L106)):*

this expression is too long, please save the values to variables
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt` line 109 at r1](https://reviewable.io/reviews/bible-translation-tools/orature/691#-NEb3Q_B8bh16GdCHB9v:-NEbJIQI1oRbq2eqTHlK:b-896fix) ([raw file](https://github.com/bible-translation-tools/orature/blob/65dc58dd56eba1ec9ac9ec8ad63b909dc7493516/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/ProjectExporter.kt#L109)):*

is it possible to transform this rx stream to flow naturally rather than have nested blocking calls?
</blockquote></details>

Done.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/691#-:-NEqdRECCTXFVTikVGes:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

PR #704, opened 2023-02-01
Fix: navigating Home while recorder is active = crash

AnonymousWalker:
When we are recording (either chapter or verse/chunk), click on "Project" breadcrumb will navigate home and cause a crash.

The problem was that the view model was trying to update the selected.txt file, after the plugin returns SUCCESS, using the "activeWorkbookProperty". This property is set to null when HomePage is docked. Thus, we have a null pointer exception

This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/704)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/704#-:-NNMQ1VU2ffy0m3zta8-:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved

PR #706, opened 2023-02-10
Narration demo

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/706)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/706)* status: 0 of 30 files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/data/primitives/Verse.kt` line 3 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/706#-NNwrp6vCkRmjVvsTGL8:-NNwrp6vCkRmjVvsTGL9:b-a60omk) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/144eec2c9144db08014e6350efce4c6a49845726/common/src/main/kotlin/org/wycliffeassociates/otter/common/data/primitives/Verse.kt#L3)):*
> ```Kotlin
> package org.wycliffeassociates.otter.common.data.primitives
> 
> data class Verse(
> ```

I can understand that this particular object might be useful specifically for a UI control, however, it should be private to that package and not a common data primitive. The reason for this is that we can create a million different data types that rearrange data in a convenient way to suit our needs for that particular domain, but that exposing them to a public package creates excessive bloat, confusion, and frustration as smaller views into such data (like this one) lack enough data to sufficiently map to other data types.  

Content (or, more likely Chunk) should be used as the data primitive that every package communicates over- and both have more than the data contained within this data class.

Now, perhaps your Verse type is useful within a UI control and forms a one directional data flow from the Chunk/Content to a View, and nothing needs to map it BACK to a Chunk/Content for processing somewhere else in the app. If that's the case, it can be package private, and should not be leaked outside of the UI layer.

As a similar example, think of Entities. It may seem unnecessary/bloat/frustrating to work with Entities, such as ContentEntity as opposed to Content. It is true they are frustrating to work with! Some objects to be useful for the domain layer really need other objects; perhaps a list of takes, or resource metadata. 

(By "Domain" I am referring to the following definition: "a specified sphere of activity or knowledge.")

So, when we try to write "use cases," these are things the user wants to do, within the "domain" of oral bible translation. Our domain classes try to model this as best as possible so that the operations on these objects naturally align with what the user is using the software to do.

The problem is, sometimes we enter a "domain" that is more tech-y. When we want data to persist, specifically in a relational database (sqlite specifically), we've entered the "domain" of a relational database, and things have to be oriented in that particular way. This adds constraints of data having to be stored in tables, of which references to other objects need to be referenced not by object pointers but foreign key relationships. 

We'd like the code we write, and the data structures we design, to suit the domain space of Bible translation rather than be shaped instead by the needs of our choice of database! As a result, private to the persistence layer (which is implemented specifically in the jvm package) we find Entities. Our repositories form the door from the domain layer to the persistence layer. Domain objects go in, and domain objects come back out. While inside the persistence layer, the database needs to work on Entities, which directly map to tables in the database. So, we first map domain objects to entities. Mapping often requires access to multiple database tables to translate object references to foreign key ids. Similarly, the same is done when data is sent back to the domain layer from the database- the entities need to be mapped from tabular data with ids to domain objects with object references.

Ultimately, the point is, these layers of abstraction exist to make working within the different "domain" spaces more natural to their needs. Data in the persistence layer is database-like, and data in the domain layer is more domain like. These almost form the languages and dialects within these realms, but the overarching language all layers should communicate across each other with should be the domain objects. Some object mappers for entities requiring multiple data access objects (daos) demonstrate how data can be lost when converting between data types, and how difficult it can be to restore that data.


All that to say- we need to be intentional about the data types we introduce and where they are introduced. Ambiguity can cause developer confusion when one of us goes to work on a new task and looks through data types and structures only to find multiple types that seem to be trying to fulfill the same purpose. If this is intended for use in views, map to it from Chunk objects in a ViewModel, and keep this type private to the package it's used in.

jsarabia (commented):
This will take me some time to fully give feedback, but so far:

I like the start so far: the ability to change theme from a dropdown is especially helpful! Thanks for starting to add some existing controls as well!

I know you're still refactoring; so some of this you might already be intending to do or in the process of doing, but:
We don't want Narration to be a control. It's a really complex page with a lot of interaction between controls (and a lot of underlying functionality we haven't engineered yet! We'll really end up rewriting Narration, not just rearranging the UI), so right now I just wanted to focus on getting all the individual pieces out there.

Each control should have a separate page, like button and combobox does.

The Narration Recording Card should be one control
The Narration Recording List which contains Narration Recording Cards should be another control
The Narratable Text Card should be one control
The Narratable Text Card List should be another control
The Resume Narratable Item should be another control

(I'm not saying these should be the names, though giving some suggestion to remove "verse" from it, keeping things generic and open to expanding functionality or at least continuing support for things such as OBS (verse is a "subclass" of a more generic term) even adding recording sections for things like front matter or titles gives a reason to not use specific words like "verse" unless necessary, and leaves open the possibility of more specific subclassing if/when needed).

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/706#-:-NNx1mAr8mX3zwRbEe8g:bezt26v)* status: 0 of 30 files reviewed, 1 unresolved discussion (waiting on @mXaln)

jsarabia (changes requested):
Okay! So- Excellent job! The controls are really well designed; while it would be nice to have them broken up in the demo app, I see that in the code they are separated out well, and while I wasn't anticipating a lot of the complex interaction yet, you have a lot of that done too and it works well! Great work!

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/706#-:-NNxIUTeCRNm9MykQtbS:b-7jvc97)* status: 0 of 30 files reviewed, 2 unresolved discussions (waiting on @mXaln)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/narration/NarrationRecordItem.kt` line 55 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/706#-NNxHiv62HUUw3iGjyzI:-NNxHiv62HUUw3iGjyzJ:bnxg9gg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/144eec2c9144db08014e6350efce4c6a49845726/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/narration/NarrationRecordItem.kt#L55)):*
> ```Kotlin
>                     graphic = FontIcon(MaterialDesign.MDI_MICROPHONE)
>                     onActionProperty().bind(onRecordAgainActionProperty)
>                 }
> ```

perhaps the menu items should be provided via a list property?

PR #708, opened 2023-02-21
Refactor import

AnonymousWalker:
Redo architecture for import

- supports import by chapters from an ongoing project

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)


jsarabia (changes requested):
It would be good to document the public APIs, as documentation is best contributed to with new code, rather than trying to add it retroactively.

Additionally, given the significance of import export, the tests for each case and edge cases should be rather extensive.

> Reviewed 4 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-:-NP-1QK04l4jshyrsghi:bq5t4dk)* status: 4 of 29 files reviewed, 4 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ImportProjectUseCase.kt` line 39 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOyfU7t4-BDCgX2z3WM:-NOyfU7t4-BDCgX2z3WN:binpbm) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ImportProjectUseCase.kt#L39)):*
> ```Kotlin
>     }
> 
>     fun isRCAlreadyImported(file: File): Boolean {
> ```

RC is a specific implementation, not the generic term (as... generic sounding of a name as it is...) so "isAlreadyImported" would be better (or isProjectImported)
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectFormatIdentifier.kt` line 25 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOyzLr357B23IwHozNk:-NOyzLr357B23IwHozNl:b-vznbh8) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectFormatIdentifier.kt#L25)):*
> ```Kotlin
>                 ProjectFormat.RESOURCE_CONTAINER
>             }
>             file.isDirectory -> {
> ```

with both cases doing the same thing, this could be 
OratureFileFormat.isSupported(file.extension), file.isDirectory ->
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 23 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOzs0HS4g0-Lt90VcyA:-NOzs0HS4g0-Lt90VcyB:bnrn9l) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L23)):*
> ```Kotlin
> 
>     @Inject
>     lateinit var deleteProvider: Provider<DeleteResourceContainer>
> ```

if you are already injecting the constructor of this class, is there a reason why you're doing field injection too?
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ImportOptions.kt` line 3 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOzuLBd6YFKXbjH2F8X:-NOzuLBd6YFKXbjH2F8Y:b6gb4vm) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ImportOptions.kt#L3)):*
> ```Kotlin
> package org.wycliffeassociates.otter.common.domain.project.importer
> 
> data class ImportOptions(
> ```

This effectively defines a public API, could you document what these values are to be used for?

AnonymousWalker (commented):
Sure. That's what I'm planning to add (tests & documentations) before having it ready to merge. Now is a draft for architectural design and implementation before I add the tests.

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-:-NP3xNiTCWm67XIOTo09:b-mxxdq3)* status: 4 of 29 files reviewed, 4 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ImportProjectUseCase.kt` line 39 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOyfU7t4-BDCgX2z3WM:-NP3xld03GQfcO1HNS1p:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ImportProjectUseCase.kt#L39)):*

RC is a specific implementation, not the generic term (as... generic sounding of a name as it is...) so "isAlreadyImported" would be better (or isProjectImported)
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectFormatIdentifier.kt` line 25 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOyzLr357B23IwHozNk:-NP3xl5_2qZ9m4O8gqm9:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectFormatIdentifier.kt#L25)):*

with both cases doing the same thing, this could be 
OratureFileFormat.isSupported(file.extension), file.isDirectory ->
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 23 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NOzs0HS4g0-Lt90VcyA:-NP3y4RNE_UTuoXybjXs:b-m16fax) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/8c05259bb07156c22dad0d76cfc4c419fa9800b6/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L23)):*

if you are already injecting the constructor of this class, is there a reason why you're doing field injection too?
</blockquote></details>

Nice catch, thanks!

jsarabia (changes requested):
> Reviewed 2 of 24 files at r4, 1 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 3 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 36 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPsRusD4sE1t5LLyFWN:-NPsRusD4sE1t5LLyFWO:b-jmh162) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L36)):*
> ```Kotlin
>         ResourceContainer.load(file).use { rc ->
>             sameVersion = (rc.manifest.dublinCore.version == existingSource.version)
>         }
> ```

should you make sure that the existing source actually exists at the path specified and that there wasn't just a dangling reference in the database?
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 49 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPsQCEgCdWqpGLcqauZ:-NPsQCEgCdWqpGLcqau_:b-5oir6l) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L49)):*
> ```Kotlin
>                 !confirmDelete -> Single.just(ImportResult.ABORTED)
>                 deleteUseCase.deleteSync(existingSource.path) == DeleteResult.SUCCESS -> {
>                     super.passToNextImporter(file, callback, options)
> ```

This is why I don't like implementing this as a linked list and prefer to have each importer in a list such that the chain is iterates over a list and returns whether the object was processed or not. 

Order can matter, but you would then only care about the order at the top level object that composes all importers and iterates over the list, and the list can be reordered easily in that file. You can't get around order mattering when multiple importers can take the same input correctly, but at least you can visually see them all in one place.

Here, you introduce a dependency on locality in two places- at the top level, where the chain is set up, but also within the chain themselves. I can't look at the top and know that for this importer to work it HAS TO be followed by the importer to import a new project because all this object did was delete the existing one. So when I'm ordering things at the top level, I might not realize this has such a dependency.

This is also just a fundamentally bad philosophy- the function should be correct within itself. It has a dependency- a SPECIFIC one, so give it the specific dependency, and don't rely on chance.

In short, inject the proper Importer that finishes the task if you want to reuse that code. Don't use a linked list where each importer is tightly coupled to the chain of responsibility implementer.

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 3 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 49 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPsQCEgCdWqpGLcqauZ:-NPwf6oI52dcIy6egd9F:b-funpww) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L49)):*

This is why I don't like implementing this as a linked list and prefer to have each importer in a list such that the chain is iterates over a list and returns whether the object was processed or not. 

Order can matter, but you would then only care about the order at the top level object that composes all importers and iterates over the list, and the list can be reordered easily in that file. You can't get around order mattering when multiple importers can take the same input correctly, but at least you can visually see them all in one place.

Here, you introduce a dependency on locality in two places- at the top level, where the chain is set up, but also within the chain themselves. I can't look at the top and know that for this importer to work it HAS TO be followed by the importer to import a new project because all this object did was delete the existing one. So when I'm ordering things at the top level, I might not realize this has such a dependency.

This is also just a fundamentally bad philosophy- the function should be correct within itself. It has a dependency- a SPECIFIC one, so give it the specific dependency, and don't rely on chance.

In short, inject the proper Importer that finishes the task if you want to reuse that code. Don't use a linked list where each importer is tightly coupled to the chain of responsibility implementer. 
</blockquote></details>

I see your point with the loop that we used in [MAUI code](https://github.com/Bible-Translation-Tools/maui/blob/dev/common/src/main/kotlin/org/bibletranslationtools/maui/common/usecases/FileProcessingRouter.kt). However, there are a few things that come back with some benefits using the "singly linked list":

With locality, there is actually only one place where the chain is set up and we can see the whole picture from there. The place is currently in the `RCImporterFactory`
```
val importer1 = ongoingProjectImporter.get()
val importer2 = existingProjectImporter.get()
val importer3 = newSourceImporter.get()

importer1.setNext(importer2)
importer2.setNext(importer3)

(return) importer1
```

This is equivalent to the for loop that you propose to replace with. However, we may have other factories later and each is supposed to produce a single Importer as its product. Using a loop may end up pushing the setup code to the Use Case layer and thus the client may not simply use Importer.import() anymore but rather some setup wrappers that contain the loop setup.

Also, using a loop with the Importers requires a new return value - ` ImportResult.UNHANDLED ` and `.blockingGet()` call to check if the importer has "consumed" the request or not.

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 4 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/MediaMerge.kt` line 34 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPxTOooB0NIChbfte87:-NPxTOooB0NIChbfte88:b-laxjw9) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/MediaMerge.kt#L34)):*
> ```Kotlin
> class MediaMerge @Inject constructor(val directoryProvider: IDirectoryProvider) {
>     private lateinit var fromRC: ResourceContainer
>     private lateinit var toRC: ResourceContainer
> ```

why not remove these entirely, pass them into all private methods, and make this purely functional.

MediaMerge can then be an object as it would have no internal state, effectively becoming just a pure function.
___
*[`jvm/workbookapp/src/integration-test/kotlin/integrationtest/projects/DatabaseEnvironment.kt` line 57 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPxSyEGAXIK1LQHzH4U:-NPxSyEGAXIK1LQHzH4V:bk51m2z) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/jvm/workbookapp/src/integration-test/kotlin/integrationtest/projects/DatabaseEnvironment.kt#L57)):*
> ```Kotlin
>     fun import(rcFile: String, importAsStream: Boolean = false, unzip: Boolean = false): DatabaseEnvironment {
>         if (importAsStream) {
>             val tempFile = directoryProvider.createTempFile("db-env-import-test", ".zip")
> ```

this doesn't appear to be deleted

(it will if you run it in CI, but locally it won't be deleted)

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 4 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/MediaMerge.kt` line 34 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPxTOooB0NIChbfte87:-NQ0udzf09a74nXxGABl:brnamd8) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/resourcecontainer/projectimportexport/MediaMerge.kt#L34)):*

why not remove these entirely, pass them into all private methods, and make this purely functional.

MediaMerge can then be an object as it would have no internal state, effectively becoming just a pure function.
</blockquote></details>

I've refactored it to an object as the number of private methods are significantly large, it would be better to have something wrapped around

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 4 unresolved discussions (waiting on @jsarabia)

___
*[`jvm/workbookapp/src/integration-test/kotlin/integrationtest/projects/DatabaseEnvironment.kt` line 57 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPxSyEGAXIK1LQHzH4U:-NQ0ux5H0RUfAWMzbHn7:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/jvm/workbookapp/src/integration-test/kotlin/integrationtest/projects/DatabaseEnvironment.kt#L57)):*

this doesn't appear to be deleted

(it will if you run it in CI, but locally it won't be deleted)
</blockquote></details>

Done.

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 4 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 36 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPsRusD4sE1t5LLyFWN:-NQ0vU6l0ooNyrKitcDg:b2zv8mi) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L36)):*

should you make sure that the existing source actually exists at the path specified and that there wasn't just a dangling reference in the database?
</blockquote></details>

if the database contains such outdated reference to the source, which is invalid data, it's a serious problem which can affect everything that touches the source. This particular instance is just one of many places, we may consider whether it is necessary to apply this defensive strategy in all places (?)

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708)* status: 5 of 37 files reviewed, 4 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt` line 36 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NPsRusD4sE1t5LLyFWN:-NQ0x9mA0QGsaazM_mx8:b-ce0jx7) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/955724681ea9c64f5b1462c95c63212330c7d719/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/ExistingSourceImporter.kt#L36)):*

if the database contains such outdated reference to the source, which is invalid data, it's a serious problem which can affect everything that touches the source. This particular instance is just one of many places, we may consider whether it is necessary to apply this defensive strategy in all places (?)
</blockquote></details>

*or better address the issue with the database being out-of-sync

jsarabia (approved):
:lgtm:

> Reviewed 3 of 29 files at r1, 3 of 24 files at r4, 1 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-:-NQRQS7f1r9tXrnTJj22:bnfp4nl)* status: 12 of 37 files reviewed, 3 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/OngoingProjectImporter.kt` line 101 at r7](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NQR5CSNA_2eOZ4c52hR:-NQR5CSNA_2eOZ4c52hS:b-i6dwwm) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dd0790bb41193bf68cc5f23a66d0de2b7b64a248/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/OngoingProjectImporter.kt#L101)):*
> ```Kotlin
>                     if (!filterProvided) {
>                         return@flatMap Single.just(ImportResult.ABORTED)
>                     }                }
> ```

formatting

AnonymousWalker:
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/OngoingProjectImporter.kt` line 101 at r7](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/708#-NQR5CSNA_2eOZ4c52hR:-NQSW_Bt6gc3VJNMgzAU:b-2w4l64) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dd0790bb41193bf68cc5f23a66d0de2b7b64a248/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/importer/OngoingProjectImporter.kt#L101)):*

formatting
</blockquote></details>

thanks! I fixed in the export PR

PR #718, opened 2023-03-09
Created Pcm classes and Pcm converter

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @mXaln)

___
*[`common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt` line 114 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQAovUd9w2pirR2q84d:-NQAovUd9w2pirR2q84e:bau88md) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bfa6ca3d2677dbc57733e9e7132bdd9c59cfbb7d/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt#L114)):*
> ```Kotlin
> 
>     fun reader(start: Int? = null, end: Int? = null): AudioFileReader {
>         return strategy.reader(start, end)
> ```

There could have been a reason to create readers as new objects rather than share one per AudioFile object.

If this change is arbitrary to attempt to clean up code, please revert the change, as this code was likely intentionally written this way. If I had to guess, the reason is that the strategy could shift pointers around when reading out metadata that could collide with readers trying to read. I don't know, the code here is old, so I'm going to be hesitant.

If the context and use case is fully understood such that you can demonstrate why this is a safe change, please explain.

Same applies to the writer.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt` line 81 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQAtCo7-tlAOlFnn0ms:-NQAtCo7-tlAOlFnn0mt:byo2m05) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bfa6ca3d2677dbc57733e9e7132bdd9c59cfbb7d/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt#L81)):*
> ```Kotlin
>             }
> 
>             cleanup(pcmReader, wavWriter)
> ```

this should probably be in a try/finally

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt` line 114 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQAovUd9w2pirR2q84d:-NQOnPr6DIR43Fes-yRc:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bfa6ca3d2677dbc57733e9e7132bdd9c59cfbb7d/common/audio/src/main/kotlin/org/wycliffeassociates/otter/common/audio/AudioFile.kt#L114)):*

There could have been a reason to create readers as new objects rather than share one per AudioFile object.

If this change is arbitrary to attempt to clean up code, please revert the change, as this code was likely intentionally written this way. If I had to guess, the reason is that the strategy could shift pointers around when reading out metadata that could collide with readers trying to read. I don't know, the code here is old, so I'm going to be hesitant.

If the context and use case is fully understood such that you can demonstrate why this is a safe change, please explain.

Same applies to the writer.
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt` line 81 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQAtCo7-tlAOlFnn0ms:-NQOnMqp9LNp96_0kUXv:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bfa6ca3d2677dbc57733e9e7132bdd9c59cfbb7d/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt#L81)):*

this should probably be in a try/finally
</blockquote></details>

Done.

common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt
```diff
@@ -39,4 +46,51 @@ class AudioConverter @Inject constructor() {
...
+
+            try {
+                pcmReader.open()
+                val buffer = ByteArray(10240)
+                while (pcmReader.hasRemaining()) {
+                    val written = pcmReader.getPcmBuffer(buffer)
+                    wavWriter.write(buffer, 0, written)
+                }
+            } finally {
+                cleanup(pcmReader, wavWriter)
```
jsarabia: rather than try/finally, it would probably be better to modify things to allow for .use { }

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt` line 86 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQ5ugvDZfmSQSe0NwM5-r2-86:-NR3KYJT7kk3cF_eBbPS:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/e5a0a8c621edf7fef7df94fadffce0ec65745dc8/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt#L86)):*

rather than try/finally, it would probably be better to modify things to allow for .use { }
</blockquote></details>

Done.

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt` line 86 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQ5ugvDZfmSQSe0NwM5-r2-86:-NR4DQRB2sTuzfJfEvKd:bfnsj72) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/e5a0a8c621edf7fef7df94fadffce0ec65745dc8/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt#L86)):*

Done.
</blockquote></details>

The reader needs it as well; and release should be called in the overridden close method. (reading could throw an exception here as well, so it should be done in a try/finally, the reason to use the use block is because things like close can also throw exceptions, so the use block should implement the try/finally correctly- it's worth looking at the source code for the use function!)

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt` line 86 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-NQ5ugvDZfmSQSe0NwM5-r2-86:-NR89S-8BdN7wfM60DOv:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/e5a0a8c621edf7fef7df94fadffce0ec65745dc8/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioConverter.kt#L86)):*

The reader needs it as well; and release should be called in the overridden close method. (reading could throw an exception here as well, so it should be done in a try/finally, the reason to use the use block is because things like close can also throw exceptions, so the use block should implement the try/finally correctly- it's worth looking at the source code for the use function!)
</blockquote></details>

Done.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/718#-:-NRDQs6o0RjyWysnY3Vf:bnfp4nl)* status: 0 of 15 files reviewed, all discussions resolved

PR #720, opened 2023-03-13
Refactor export

AnonymousWalker:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 30 files reviewed, 4 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/Utils.kt` line 13 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgbLWQ3da8Bw5ipYWl:-NQgbLWQ3da8Bw5ipYWm:b-cs1ads) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/Utils.kt#L13)):*
> ```Kotlin
>     val extensionDelim = """\."""
>     Pattern.compile(chapter + verse + sort + type + take + extensionDelim)
> }
> ```

maybe this belongs in the FileNamer in the companion object?
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/IDirectoryExporter.kt` line 4 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgaPrKAnr1oW5R9oj8:-NQgaPrKAnr1oW5R9oj9:b-gisn2q) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/IDirectoryExporter.kt#L4)):*
> ```Kotlin
> 
> interface IDirectoryExporter : ProjectExporter {
> }
> ```

the { } are unnecessary
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporter.kt` line 8 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgjgSh32HhwcBu6ujg:-NQgjgSh32HhwcBu6ujh:b4az8ra) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporter.kt#L8)):*
> ```Kotlin
> import java.io.File
> 
> interface ProjectExporter {
> ```

IProjectExporter
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporterCallback.kt` line 4 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgaiHc-qsZPtI84tq3:-NQgaiHc-qsZPtI84tq4:b-gisn2q) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporterCallback.kt#L4)):*
> ```Kotlin
> 
> interface ProjectExporterCallback {
> }
> ```

the { } are unnecessary

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 30 files reviewed, 6 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt` line 6 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgn_2vD5Sx_7m7F7Wj:-NQgn_2vD5Sx_7m7F7Wk:butfltp) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt#L6)):*
> ```Kotlin
> import org.wycliffeassociates.otter.common.data.primitives.ResourceMetadata
> 
> open class ProjectMetadata(
> ```

If this is made from the Resource Metadata, and only contains values in the Resource Metadata... it feels like this is just clutter and we'd prefer to just pass around the Resource Metadata
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt` line 127 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgnoj_EeHqX2G9k8sK:-NQgnoj_EeHqX2G9k8sL:b87sdcz) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt#L127)):*
> ```Kotlin
>     private fun chapterFilter(chapter: Chapter, options: ExportOptions?): Boolean {
>         val included = options?.let { chapter.sort in it.chapters } ?: true
>         return included && chapter.audio.selected.value?.value != null
> ```

too much implementation details of the Chapter leaking out here;

it should be chapter.hasSelectedAudio()

and the chapter should have a method that makes those calls. Then if somehow the implementation of things within the chapter changes, this code won't matter

AnonymousWalker:
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt` line 6 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgn_2vD5Sx_7m7F7Wj:-NR3bD02AJWW-0Cdhwdt:b-ffssj7) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt#L6)):*

If this is made from the Resource Metadata, and only contains values in the Resource Metadata... it feels like this is just clutter and we'd prefer to just pass around the Resource Metadata
</blockquote></details>

actually, I was trying to do the opposite. ResourceMetadata strongly associates with resource container, particularly. I attempted to "abstract" from that class by having the ProjectMetadata as a more universal one. The ability to construct the object from a ResourceMetadata instance could be done in a separate mapper function, for example`toProjectMetadata()`,  but I found it convenient to use a secondary constructor instead.

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 30 files reviewed, 6 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/Utils.kt` line 13 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgbLWQ3da8Bw5ipYWl:-NR4-JVW174D7oxwsFdO:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/Utils.kt#L13)):*

maybe this belongs in the FileNamer in the companion object?
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/IDirectoryExporter.kt` line 4 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgaPrKAnr1oW5R9oj8:-NR4-I-b4EsTST1Q0hfT:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/IDirectoryExporter.kt#L4)):*

the { } are unnecessary
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporter.kt` line 8 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgjgSh32HhwcBu6ujg:-NR4-IEi8u5lakJugzRT:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporter.kt#L8)):*

IProjectExporter
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporterCallback.kt` line 4 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgaiHc-qsZPtI84tq3:-NR4-Hjk10S72XGjgV5A:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/ProjectExporterCallback.kt#L4)):*

the { } are unnecessary
</blockquote></details>

Done.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt` line 127 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgnoj_EeHqX2G9k8sK:-NR3zfMp4hpY_vMKNlph:bgsvy17) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt#L127)):*

too much implementation details of the Chapter leaking out here;

it should be chapter.hasSelectedAudio()

and the chapter should have a method that makes those calls. Then if somehow the implementation of things within the chapter changes, this code won't matter
</blockquote></details>

great catch! Thanks

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 32 files reviewed, 6 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt` line 6 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgn_2vD5Sx_7m7F7Wj:-NR9ZDs53J2neeRLf05x:bseltd1) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt#L6)):*

actually, I was trying to do the opposite. ResourceMetadata strongly associates with resource container, particularly. I attempted to "abstract" from that class by having the ProjectMetadata as a more universal one. The ability to construct the object from a ResourceMetadata instance could be done in a separate mapper function, for example`toProjectMetadata()`,  but I found it convenient to use a secondary constructor instead.
</blockquote></details>

I understand the sentiment here, so allow me to answer two different aspects of this.

__First__ is in regards to the desire to abstract because of the identification of other potential container formats. Yes, you're coming at this with the right attitude. Yes the resource metadata does look very, very similar to the manifest of a resource container. As hard as it might be to believe, it was my intention that the resource metadata WOULD be separate from a resource container. The idea would be that other containers would map their metadata to this- and all containers will have metadata. All known container formats we consider at the moment (epub, RC, burrito) use dublin core inspired metadata. All of these fields apply. Some are perhaps irrelevant- but only the subject and description come to mind.

conformsTo is the versioning of the container format, all will have this, and it could be important for using the right library for reading the container

format is the mimetype of the primary data format in the container (markdown, usfm, html, whatever) This is important for selecting the right content reader.

identifier is the collection slug of a resource, ulb or ayt for example, and this will be important regardless of container format.

issued, modified dates, and version could all be relevant for knowing the particular version of a source resource.

publisher and creators are used to further differentiate between versions of resources, and we use this to distinguish between, say, unfoldingWord's version of a ULB (now they named theirs ULT which is nice, but there's the potential for collisions, of which this used to be an example) and as such we actively use this in project paths for the reason of potential collisions.

All content regardless of container format has a license

All containers will have a path to where the container is located

All containers have a "type" of content. RCs call this type, and we use the same vocabulary; burrito uses "flavor," but ultimately this is important for determining fundamentally what the container can be used for. Is it a book that can be translated or narrated? Or instead is it a helper resource like notes or questions which kind of depends on some other resource?

__Second__ This is really the wrong side of the equation to suddenly care about this. Import time is when a container format needs to map to this. Once the data is mapped to it, Orature is free to use this info and does. This class belongs to Orature as a data type after all, it is not imported from another library. Just because it closely resembles something else (because that something else was our primary intended use case, well understood, and as such the abstraction wasn't trying to balance between multiple implementations) does not mean that it is an actual tight coupling to that thing. So, if things are imported, it logically follows that this data exists. Restricting down to a smaller subset is just code bloat. You have this class, you can retrieve it out of the repository for a resource when you want to export something. If you don't choose to use all the fields when you export, then don't worry that it contains more data than you need. Taking it, in order to use this to map to a class to remove unnecessary fields is an extra step, adds a source file, and if in the future you find you want that data back... well the refactor gets to be deeper and more extensive because having more data than you need isn't a problem, but having less data than you need is.

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 32 files reviewed, 2 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt` line 56 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NRDObpR7RoswrYCtaff:-NRDObpR7RoswrYCtafg:b-1nvqc7) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/04de40ab3eaf7c156b32461dc3ee0afa97c33e61/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt#L56)):*
> ```Kotlin
>             workbook.target.toCollection()
>         )
>         val isBook = projectMetadata.resourceSlug == workbook.target.resourceMetadata.identifier
> ```

We could stand to clean up some code here. 

We effectively got rid of support for Resources with the changes to Chunking. We also felt that the way resources worked before made things like navigation and export challenging and confusing- so that if we were to add resources back in the future, we would do them as standalone, sovereign projects like other books.

As such, things like checking if the rc metadata is a book doesn't make sense?

As such, we can probably remove this else statement, as well as simplify the input parameters because the workbook has the metadata of the target book

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 32 files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt` line 56 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NRDObpR7RoswrYCtaff:-NRZoPpU2Iej9Dl9gug9:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/04de40ab3eaf7c156b32461dc3ee0afa97c33e61/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/AudioProjectExporter.kt#L56)):*

We could stand to clean up some code here. 

We effectively got rid of support for Resources with the changes to Chunking. We also felt that the way resources worked before made things like navigation and export challenging and confusing- so that if we were to add resources back in the future, we would do them as standalone, sovereign projects like other books.

As such, things like checking if the rc metadata is a book doesn't make sense?

As such, we can probably remove this else statement, as well as simplify the input parameters because the workbook has the metadata of the target book
</blockquote></details>

Done.

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 0 of 32 files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt` line 6 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NQgn_2vD5Sx_7m7F7Wj:-NRZoXq_3E6g6GyCp_0C:b-k9luk0) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0b744b71b843fb4d0320c55cc431886c4406b756/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/ProjectMetadata.kt#L6)):*

I understand the sentiment here, so allow me to answer two different aspects of this.

*First* is in regards to the desire to abstract because of the identification of other potential container formats. Yes, you're coming at this with the right attitude. Yes the resource metadata does look very, very similar to the manifest of a resource container. As hard as it might be to believe, it was my intention that the resource metadata WOULD be separate from a resource container. The idea would be that other containers would map their metadata to this- and all containers will have metadata. All known container formats we consider at the moment (epub, RC, burrito) use dublin core inspired metadata. All of these fields apply. Some are perhaps irrelevant- but only the subject and description come to mind.

conformsTo is the versioning of the container format, all will have this, and it could be important for using the right library for reading the container

format is the mimetype of the primary data format in the container (markdown, usfm, html, whatever) This is important for selecting the right content reader.

identifier is the collection slug of a resource, ulb or ayt for example, and this will be important regardless of container format.

issued, modified dates, and version could all be relevant for knowing the particular version of a source resource.

publisher and creators are used to further differentiate between versions of resources, and we use this to distinguish between, say, unfoldingWord's version of a ULB (now they named theirs ULT which is nice, but there's the potential for collisions, of which this used to be an example) and as such we actively use this in project paths for the reason of potential collisions.

All content regardless of container format has a license

All containers will have a path to where the container is located

All containers have a "type" of content. RCs call this type, and we use the same vocabulary; burrito uses "flavor," but ultimately this is important for determining fundamentally what the container can be used for. Is it a book that can be translated or narrated? Or instead is it a helper resource like notes or questions which kind of depends on some other resource?

*Second* This is really the wrong side of the equation to suddenly care about this. Import time is when a container format needs to map to this. Once the data is mapped to it, Orature is free to use this info and does. This class belongs to Orature as a data type after all, it is not imported from another library. Just because it closely resembles something else (because that something else was our primary intended use case, well understood, and as such the abstraction wasn't trying to balance between multiple implementations) does not mean that it is an actual tight coupling to that thing. So, if things are imported, it logically follows that this data exists. Restricting down to a smaller subset is just code bloat. You have this class, you can retrieve it out of the repository for a resource when you want to export something. If you don't choose to use all the fields when you export, then don't worry that it contains more data than you need. Taking it, in order to use this to map to a class to remove unnecessary fields is an extra step, adds a source file, and if in the future you find you want that data back... well the refactor gets to be deeper and more extensive because having more data than you need isn't a problem, but having less data than you need is.


</blockquote></details>

Great! Thanks for clearing my concern

jsarabia (changes requested):
> Reviewed 2 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 2 of 37 files reviewed, 2 unresolved discussions (waiting on @AnonymousWalker)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/resourcecontainer/RCProjectExporter.kt` line 59 at r4](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NRhMbUw0OCzkN_hRqGa:-NRhMbUw0OCzkN_hRqGb:bgwhl8z) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/cb378a28cfb817047c0135a912de17cbe91e6222/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/resourcecontainer/RCProjectExporter.kt#L59)):*
> ```Kotlin
>     private val compositeDisposable = CompositeDisposable()
> 
>     protected fun getProjectFileAccessor(workbook: Workbook): ProjectFilesAccessor {
> ```

honestly, this could be a method of the workbook

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720)* status: 2 of 37 files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/resourcecontainer/RCProjectExporter.kt` line 59 at r4](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/720#-NRhMbUw0OCzkN_hRqGa:-NRhUH13CSkWlZWhTQRS:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/cb378a28cfb817047c0135a912de17cbe91e6222/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/project/exporter/resourcecontainer/RCProjectExporter.kt#L59)):*

honestly, this could be a method of the workbook
</blockquote></details>

Done.

PR #745, opened 2023-05-03
Create Translation Wizard

AnonymousWalker:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745)


mXaln (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745)* status: all files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt` line 46 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745#-NV-pJYd1x3watIbStz7:-NV-pJYd1x3watIbStz8:bjujc5) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dec2af42e6a0b9d4e99cb549a412d06acadf4676/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt#L46)):*
> ```Kotlin
> 
>     fun setOnSelectAction(op: () -> Unit) {
>         onSelectAction = op
> ```

I would use property Action here instead of just lambda function, because lambdas never get garbage collected. But properties use binding which uses WeakListener that could be garbage collected at some point. 
ex. `onSelectActionProperty.set(op)`

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745)* status: all files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt` line 46 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745#-NV-pJYd1x3watIbStz7:-NV0WFThBioRpWkV2IhJ:bokl280) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dec2af42e6a0b9d4e99cb549a412d06acadf4676/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt#L46)):*

I would use property Action here instead of just lambda function, because lambdas never get garbage collected. But properties use binding which uses WeakListener that could be garbage collected at some point. 
ex. `onSelectActionProperty.set(op)`
</blockquote></details>

I agree with the idea, however, this particular view is a wizard that contains 3 buttons (static), so memory leak may not be a concern here. Would it be better to keep the simplicity of this lambda instead of using EventHandler<ActionEvent>?

mXaln (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745)* status: all files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt` line 46 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745#-NV-pJYd1x3watIbStz7:-NV4s5he2JHrLxe0lAg4:b-sfbdxo) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dec2af42e6a0b9d4e99cb549a412d06acadf4676/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt#L46)):*

I agree with the idea, however, this particular view is a wizard that contains 3 buttons (static), so memory leak may not be a concern here. Would it be better to keep the simplicity of this lambda instead of using EventHandler<ActionEvent>?
</blockquote></details>

I don't have strong position on this. So if you think that simplicity is more important here, them I'm ok with it

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745)* status: all files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt` line 46 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745#-NV-pJYd1x3watIbStz7:-NVBTlE07_-QgTa-bKF-:b-jvc1sw) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dec2af42e6a0b9d4e99cb549a412d06acadf4676/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt#L46)):*

I don't have strong position on this. So if you think that simplicity is more important here, them I'm ok with it
</blockquote></details>

The event handler is better because it's too easy to write code in the lambda that causes the leak, and it's easy to forget that the underlying setAction method ISN'T implemented using properties

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745)* status: all files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt` line 46 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745#-NV-pJYd1x3watIbStz7:-NVF2fqV0GKc1nuZih-h:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dec2af42e6a0b9d4e99cb549a412d06acadf4676/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/card/TranslationTypeCard.kt#L46)):*

The event handler is better because it's too easy to write code in the lambda that causes the leak, and it's easy to forget that the underlying setAction method ISN'T implemented using properties
</blockquote></details>

Done.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/745#-:-NW2EY4AA-SzTSQUVesh:bnfp4nl)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on @mXaln)

PR #785, opened 2023-07-27
Fix delete-then-import a project

AnonymousWalker:
When deleting a project, the workbook's collection still exists because we reinsert it. As a result, the verses content are not initialized and all the verse takes from the input file will not be added during import.

This PR does two things:
- makes the deleteProjects() a hard delete for all derived collections and contents
- always attempts to derive the content (verses/chunks) under chapter-level when importing a workbook project

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/785)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/785#-:-Nahb2Kn3OXsl-YrHP-B:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved

PR #794, opened 2023-08-09
Refactor narration viewmodels

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt
@@ -8,6 +8,7 @@ import io.reactivex.subjects.PublishSubject
 import org.slf4j.LoggerFactory
 import org.wycliffeassociates.otter.common.audio.AudioFileReader
 import org.wycliffeassociates.otter.common.data.workbook.Chapter
+import org.wycliffeassociates.otter.common.data.workbook.Chunk
 import org.wycliffeassociates.otter.common.data.workbook.Workbook
 import org.wycliffeassociates.otter.common.domain.audio.OratureAudioFile
 import java.io.File
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/di/AppDependencyGraph.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/di/AppDependencyGraph.kt
@@ -49,6 +49,7 @@ import org.wycliffeassociates.otter.jvm.device.audio.AudioConnectionFactory
 import org.wycliffeassociates.otter.jvm.device.audio.AudioDeviceProvider
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.NarrationDebugApp
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.AudioWorkspaceViewModel
+import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.NarrationViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.AppInfoViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.ChunkingViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.ExportProjectViewModel
@@ -91,6 +92,7 @@ interface AppDependencyGraph {
     fun inject(viewModel: ExportChapterViewModel)
     fun inject(viewModel: ExportProjectViewModel)
     fun inject(viewModel: AudioWorkspaceViewModel)
+    fun inject(viewModel: NarrationViewModel)
     fun injectDatabase(): AppDatabase
     fun injectDirectoryProvider(): IDirectoryProvider
     fun injectAppPreferencesRepository(): IAppPreferencesRepository
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/di/AppDependencyGraph.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/di/AppDependencyGraph.kt
@@ -47,7 +47,7 @@ import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.WorkbookPageVie
 import javax.inject.Singleton
 import org.wycliffeassociates.otter.jvm.device.audio.AudioConnectionFactory
 import org.wycliffeassociates.otter.jvm.device.audio.AudioDeviceProvider
-import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.NarrationApp
+import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.NarrationDebugApp
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.AudioWorkspaceViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.AppInfoViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.ChunkingViewModel
@@ -69,7 +69,7 @@ import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.RootViewModel
 @Singleton
 interface AppDependencyGraph {
     fun inject(app: OtterApp)
-    fun inject(app: NarrationApp)
+    fun inject(app: NarrationDebugApp)
 
     fun inject(viewModel: RootViewModel)
     fun inject(viewModel: SplashScreenViewModel)
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/di/AppDependencyGraph.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/di/AppDependencyGraph.kt
@@ -48,7 +48,7 @@ import javax.inject.Singleton
 import org.wycliffeassociates.otter.jvm.device.audio.AudioConnectionFactory
 import org.wycliffeassociates.otter.jvm.device.audio.AudioDeviceProvider
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.NarrationApp
-import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.NarrationBodyViewModel
+import org.wycliffeassociates.otter.jvm.workbookapp.ui.narration.AudioWorkspaceViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.AppInfoViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.ChunkingViewModel
 import org.wycliffeassociates.otter.jvm.workbookapp.ui.viewmodel.ExportProjectViewModel
@@ -90,7 +90,7 @@ interface AppDependencyGraph {
     fun inject(viewModel: ChunkingViewModel)
     fun inject(viewModel: ExportChapterViewModel)
     fun inject(viewModel: ExportProjectViewModel)
-    fun inject(viewModel: NarrationBodyViewModel)
+    fun inject(viewModel: AudioWorkspaceViewModel)
     fun injectDatabase(): AppDatabase
     fun injectDirectoryProvider(): IDirectoryProvider
     fun injectAppPreferencesRepository(): IAppPreferencesRepository
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/AudioWorkspaceView.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/AudioWorkspaceView.kt
@@ -97,22 +97,14 @@ class AudioWorkspaceViewModel : ViewModel() {
 
     val isRecordingProperty = SimpleBooleanProperty()
     var recordedVerses = observableListOf<VerseNode>()
-
-    private val listeners = mutableListOf<ListConversionListener<*, *>>()
-
+    
     init {
         (app as IDependencyGraphProvider).dependencyGraph.inject(this)
     }
 
     fun onDock() {
         isRecordingProperty.bind(narrationViewModel.isRecordingProperty)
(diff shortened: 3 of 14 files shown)


AnonymousWalker (approved):

mXaln (changes requested):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)* status: all files reviewed, 2 unresolved discussions (waiting on @darrellcolehill and @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-NbZeV527NE4rh-9bMaO:b-fbe7w1) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*
> ```Kotlin
>         }.let { eventSubscriptions.add(it) }
> 
>         subscribe<PluginClosedEvent> {
> ```

This event will not be caught, when opening an external app, because NarrationPage will be undocked, and events are not received on undocked views. That's why I kept these subscriptions in view model, because view modes are never undocked.
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 132 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZdNmHAjFIldYxoYh0:-NbZdNmI7mIPtKuErIU-:b-s1rohf) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L132)):*
> ```Kotlin
>             }
>             .map { list ->
>                 val chapterToResume = list
> ```

The list is not nullable here, so nullcheck is unnecessary. Did you mean `list.firstOrNull { !it.hasSelectedAudio() }`?

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)* status: all files reviewed, 2 unresolved discussions (waiting on @darrellcolehill and @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-NbeEI9S0UOIr8v9vTlX:bxeanew) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*

This event will not be caught, when opening an external app, because NarrationPage will be undocked, and events are not received on undocked views. That's why I kept these subscriptions in view model, because view modes are never undocked.
</blockquote></details>

Oh, good catch. That's not good.
Since we want to clean up things, we had been adding a dock/undock method in viewmodels to call from onDock/onUndock because I can't think of another way to ensure we clean up resources like subscriptions when we move away from the page...

So, the onUndock call should then be calling the onUndock call in the viewmodel, which would clear the subscriptions. 

Sounds like maybe a better approach would not be to dock a different page when a plugin is opened, but rather have this page place the layout for a plugin being opened inside the NarrationView using a stackpane, and toggle visibility when the plugin is opened?
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 132 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZdNmHAjFIldYxoYh0:-NbeGV9s68UykZf308CX:b5ct3xl) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L132)):*

The list is not nullable here, so nullcheck is unnecessary. Did you mean `list.firstOrNull { !it.hasSelectedAudio() }`?
</blockquote></details>

ohhh, I think I had done a .find call first before first, but then realized I could just put the predicate in the first call?

Either way, yes, you're right, thanks

darrellcolehill (approved):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)* status: all files reviewed, 2 unresolved discussions (waiting on @mXaln)

mXaln (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)* status: all files reviewed, 1 unresolved discussion (waiting on @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-NboHiez5MoUfb2YgcIJ:b-1ufna3) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*

Oh, good catch. That's not good.
Since we want to clean up things, we had been adding a dock/undock method in viewmodels to call from onDock/onUndock because I can't think of another way to ensure we clean up resources like subscriptions when we move away from the page...

So, the onUndock call should then be calling the onUndock call in the viewmodel, which would clear the subscriptions. 

Sounds like maybe a better approach would not be to dock a different page when a plugin is opened, but rather have this page place the layout for a plugin being opened inside the NarrationView using a stackpane, and toggle visibility when the plugin is opened?
</blockquote></details>

Yeah, that might work better.

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)* status: 12 of 15 files reviewed, 1 unresolved discussion (waiting on @darrellcolehill and @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-Nbzrfgp5U4ta_4bwzhr:b910374) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*

Yeah, that might work better.
</blockquote></details>

I moved the subscription to the RootView and removed it from the RecordScripturePage and NarrationPage. For one, in the RecordScripture page, the subscription was not being unsubscribed, so if the user went there, and then went to Narration in the same session, both subscriptions would be active and executing.

mXaln (changes requested):
Also I found a bug in NarrationHeaderViewModel not binding to the chapter take in init.

Should add:

`chapterTakeProperty.bind(narrationViewModel.chapterTakeProperty)`

Another issue, I've just realized:

in `NarrationViewModel.getChapterList` you set chapterToResume from the first chapter that doesn't have selected audio. But that means that if there is a selected take in chapter but no narration (say, chapter 1), it will be skipped to the next one without audio (say, chapter 2). And when you switch back to the chapter 1, and try to add markers in MarkerApp, it will load chapter 2 again, and chapter 1 will not be split into narration verses. I think it should always open either first chapter or we need to have a method in Chapter class, something like `narrationComplete()`, which means all the verses recorded.

> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-:-Nc29PuH3EqtW8UadVjf:bvj73xs)* status: all files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-Nc28FRuDzf5BV_KQ9f7:b5tzdxc) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*

I moved the subscription to the RootView and removed it from the RecordScripturePage and NarrationPage. For one, in the RecordScripture page, the subscription was not being unsubscribed, so if the user went there, and then went to Narration in the same session, both subscriptions would be active and executing.
</blockquote></details>

`subscribe<ChapterReturnFromPluginEvent> { 
    narrationViewModel.onChapterReturnFromPlugin()  
}`

ChapterReturnFromPluginEvent should also move to RootView. But if we need to run some function in the  local view model, shoud we inject it in the RootView? Like here, we need to cal `narrationViewModel.onChapterReturnFromPlugin()`.
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/RootView.kt` line 52 at r3](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-Nc1b-2oEmDE-tmTndto:-Nc1b-2oEmDE-tmTndtp:b-gleoem) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/eeaeb9e9f0ed0db6a078cce2455b7a7bea4664bf/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/RootView.kt#L52)):*
> ```Kotlin
>             viewModel.externalPluginOpenedProperty.set(!it.isNative)
>         }
>         workspace.subscribe<PluginClosedEvent> {
> ```

Can we subscribe to PluginClosedEvent here once, not twice?

jsarabia (commented):
Yeah, there's a couple issues with the overall state machine at the moment =\ but I do need to get this rearrangement in for Darrell to be able to work off of (it's just going into pair-narration, not Orature-2 since we're all still working) 

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-:-Nc2QZAY46fKp6D5iwfz:b-dabkr6)* status: 13 of 15 files reviewed, 2 unresolved discussions (waiting on @darrellcolehill and @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-Nc2OybQ86eFCqeeMRyH:b-dyu0sb) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*

`subscribe<ChapterReturnFromPluginEvent> { 
    narrationViewModel.onChapterReturnFromPlugin()  
}`

ChapterReturnFromPluginEvent should also move to RootView. But if we need to run some function in the  local view model, shoud we inject it in the RootView? Like here, we need to cal `narrationViewModel.onChapterReturnFromPlugin()`.
</blockquote></details>

No, we should definitely not have the viewmodel doing anything for Narration on the RootView. For one, that would prevent the NarrationViewModel from ever getting GCed.

I guess at first guess we could set a flag when we launch a plugin from the NarrationViewModel, so that the property is set and can execute code to handle returning from a plugin in the onDock. Or some other kind of state that can be identified onDock to know what to do - maybe the plugin edited section is copied out to a particular pcm file; so its existence signifies that a plugin returned.
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/RootView.kt` line 52 at r3](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-Nc1b-2oEmDE-tmTndto:-Nc2QQPhDh7qD56GZ-eY:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/eeaeb9e9f0ed0db6a078cce2455b7a7bea4664bf/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/screens/RootView.kt#L52)):*

Can we subscribe to PluginClosedEvent here once, not twice?
</blockquote></details>

Done.

mXaln (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794)* status: all files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt` line 91 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/794#-NbZeV527NE4rh-9bMaN:-Nc2SQm_Ad3IEUounaOx:bq0ncok) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/41579cc8394ce2684ca040fa56dbcd2dc8ba337c/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationPage.kt#L91)):*

No, we should definitely not have the viewmodel doing anything for Narration on the RootView. For one, that would prevent the NarrationViewModel from ever getting GCed.

I guess at first guess we could set a flag when we launch a plugin from the NarrationViewModel, so that the property is set and can execute code to handle returning from a plugin in the onDock. Or some other kind of state that can be identified onDock to know what to do - maybe the plugin edited section is copied out to a particular pcm file; so its existence signifies that a plugin returned.
</blockquote></details>

sounds good

PR #824, opened 2023-09-13
Change pcm reader

jsarabia:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/824)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioScene.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioScene.kt
@@ -33,7 +33,7 @@ class AudioScene(
 
     val frameBuffer = FloatArray(width * 2)
 
-    fun getNarrationDrawable(location: Int): FloatArray {
+    fun getNarrationDrawable(location: Int): Pair<FloatArray, IntRange> {
         Arrays.fill(frameBuffer, 0f)
         val framesOnScreen = secondsOnScreen * recordingSampleRate
 
@@ -65,7 +65,7 @@ class AudioScene(
                 System.arraycopy(activeData, 0, frameBuffer, 0, activeData.size)
             }
         }
-        return frameBuffer
+        return Pair(frameBuffer, viewPortRange)
     }
 
     /**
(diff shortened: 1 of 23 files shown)


darrellcolehill (changes requested):
> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/824)* status: all files reviewed, 2 unresolved discussions (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt` line 262 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/824#-NeET52gEaRIUVsNUaUI:-NeET52gEaRIUVsNUaUJ:bpnbe0p) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d8c3c5e1e5a89a6a62d968b74ca4ba0c2891a148/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt#L262)):*
> ```Kotlin
> 
>             var index = lockToVerse.get()
>             index = if (index == CHAPTER_UNLOCKED) 0 else index
> ```

I agree that making a state variable for this condition, or wrapping it in a function will help later on with debugging/readability. 
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt` line 312 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/824#-NeETqz_AJtX1RFONMlk:-NeETqz_AJtX1RFONMll:bxxzov5) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d8c3c5e1e5a89a6a62d968b74ca4ba0c2891a148/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt#L312)):*
> ```Kotlin
>             return when (lockToVerse.get()) {
>                 CHAPTER_UNLOCKED -> getPcmBufferChapter(bytes)
>                 else -> getPcmBufferVerse(bytes, activeVerses[lockToVerse.get()])
> ```

This is more of a note and somewhat ties into the above comment. This may be a bit misleading because when I call getPcmBuffer, and the returned value is less than the size of my buffer, I would be tempted to think that means I have read the whole file, and not that I have just read an entire verse and that I need to continue to the next verse.

darrellcolehill (approved):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/824)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsarabia)

PR #827, opened 2023-09-18
Dh off by one fixes

darrellcolehill:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/827)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterRepresentation.kt
@@ -15,7 +15,6 @@ import org.wycliffeassociates.otter.common.domain.audio.OratureAudioFile
 import java.io.File
 import java.io.RandomAccessFile
 import java.util.concurrent.atomic.AtomicInteger
-import kotlin.math.log
 import kotlin.math.max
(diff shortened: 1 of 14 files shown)


darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/827)* status: 0 of 13 files reviewed, 1 unresolved discussion

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/VerseNode.kt` line 215 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/827#-NeibfES4UBNT7yYmv_P:-NeibfES4UBNT7yYmv_Q:box33jr) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/dd2d76a7d7fc2187fa561f02fb3bf6cc14e7325f/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/VerseNode.kt#L215)):*
> ```Kotlin
>                 //  of frames up to (not including) the absolute frame.
>                 //  I.e. if the question this function is answering is "how many frames are behind my current position?"
>                 frameOffset += absoluteFrame - sector.first
> ```

This line I removed the +1 that was previously here because it was causing the off-by-one in ChapterRepresentation.absoluteToRelative

PR #863, opened 2023-11-01
Bug 734: narration menu options disabled after save

darrellcolehill:
Updates the chapterTakeProperty when a new take is finished being created so that the UI can react accordingly.

Sets the chapterTakeProperty equal to workbookDataStore.chapter.getSelectedTake to account for if a user goes to the next chapter before the chapter take is finished being created.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/863)

### DIFF
--- a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt
+++ b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt
@@ -606,6 +606,7 @@ class NarrationViewModel : ViewModel() {
                             .subscribeOn(Schedulers.io())
                             .subscribe(
                                 {
+                                    chapterTakeProperty.set(workbookDataStore.chapter.getSelectedTake())
                                     logger.info("Created a chapter take for ${chapterTitleProperty.value}")
                                 }, { e ->
                                     logger.error(


mXaln (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/863)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @darrellcolehill and @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 609 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/863#-NioZ47T18EZXsL8WXcJ:-NioZ47T18EZXsL8WXcK:batlywu) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/5947e12e4bf9eb3758874d874f81b4d4c11707fb/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L609)):*
> ```Kotlin
>                             .subscribe(
>                                 {
>                                     chapterTakeProperty.set(workbookDataStore.chapter.getSelectedTake())
> ```

There is a problem with this approach. In the function `createChapterTake()` `insertTake(take)` is called, which inserts the take into database, and then `selectedTakeRelay` is triggered to update the database. So here on subscribe `chapter.getSelectedTake()` could return the old selected take instead of the current one. I have addressed this bug fix in https://github.com/Bible-Translation-Tools/Orature/pull/879

PR #872, opened 2023-11-06
Fix crash when rechunk - Translation

AnonymousWalker:
Race condition occurred when deleting right before adding chunks to chapter

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/872)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/Chapter.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/data/workbook/Chapter.kt
@@ -19,6 +19,7 @@
 package org.wycliffeassociates.otter.common.data.workbook
 
 import com.jakewharton.rxrelay2.BehaviorRelay
+import io.reactivex.Completable
 import io.reactivex.Maybe
 import io.reactivex.Observable
 import io.reactivex.Single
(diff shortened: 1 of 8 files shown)


jsarabia (commented):
I'm okay with changing these lambdas to return observables. However, if you're going to do that, you must refactor where they are used to be async, and not use blocking waits.

Ultimately, this is domain code, which means it should be cross platform. One of the main purposes of architecting Orature to have a separate common package from JVM is for the ability to maintain as much Android compatible code in common as possible. The Android OS will crash the app if database operations are performed on the main thread- which was the sole purpose behind me choosing to use RxJava in Orature to begin with.

I know it's frustrating, but we have to try to be as vigilant as possible to not compromise on major architectural decisions because small inclusions like this over time will cause the amount of technical debt to move to Android to increase to a point where small refactors are not possible, and substantial re-writes become necessary. I may not catch everything, but we have to try.

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/872#-:-Nij4HGA5PTPaRt8fnSY:bv8grrj)* status: 0 of 4 files reviewed, all discussions resolved

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/872#-:-NipJquM8YaopDhemwg9:bnfp4nl)* status: 0 of 8 files reviewed, all discussions resolved

PR #882, opened 2023-11-13
Sync scrollbar with waveform

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/882)

### DIFF
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt
@@ -447,6 +447,10 @@ class Narration @AssistedInject constructor(
         chapterReaderConnection.seek(location)
     }
 
+    fun getFramePosition(): Int {
+        return chapterRepresentation.absoluteToRelativeChapter(chapterReaderConnection.absoluteFramePosition)
+    }
+
     /**
      * Gets the duration of the relative chapter space in frames
      */
--- a/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt
+++ b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt
@@ -458,7 +458,7 @@ class Narration @AssistedInject constructor(
         return chapterReaderConnection.totalFrames + uncommittedRecordedFrames.get()
     }
 
-    fun getRelativeChapterLocation(): Int {
+    private fun getRelativeChapterLocation(): Int {
         return if (lockedVerseIndex != null) {
             chapterReaderConnection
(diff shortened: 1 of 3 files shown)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/882)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 450 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/882#-Nj7zsg55YN4O19Oe06m:-Nj7zsg55YN4O19Oe06n:b-gxut7j) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/529ad0f87971a2b2a423e04ca4bf49c22ef86368/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L450)):*
> ```Kotlin
>     }
> 
>     fun getFramePosition(): Int {
> ```

this function name means the same thing as the function a few functions down - getLocationInFrames

first, where we place functions matters, and if there are a bunch are related to location, then this should go next to them and not be interrupted by different functions

second, you probably want getLocationInFrames, and this function shouldn't exist

if you do need it, then some renaming of this and other methods around here needs to happen to clarify intent and when a method should be used over another. Comments are important for that, but good method naming is more important.

getTotalFrames should probably be moved adjacent to getDurationInFrames

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/882)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 450 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/882#-Nj7zsg55YN4O19Oe06m:-Nj8PacW7w9XaTukvREh:bc0or8q) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/529ad0f87971a2b2a423e04ca4bf49c22ef86368/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L450)):*

this function name means the same thing as the function a few functions down - getLocationInFrames

first, where we place functions matters, and if there are a bunch are related to location, then this should go next to them and not be interrupted by different functions

second, you probably want getLocationInFrames, and this function shouldn't exist

if you do need it, then some renaming of this and other methods around here needs to happen to clarify intent and when a method should be used over another. Comments are important for that, but good method naming is more important.

getTotalFrames should probably be moved adjacent to getDurationInFrames
</blockquote></details>

Initially I used a different approach to solve the problem and I needed that function. But now as I changed the way to update the scrollbar, that function is not needed.

PR #909, opened 2023-11-29
Correctly calculate waveform width by using Double (OT-770)

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/909)


AnonymousWalker (approved):
Great work! I applied the fix to verse marker app, too

PR #928, opened 2023-12-07
Create chapter take when re-record and edit

AnonymousWalker:
#narration
- When chapter is completed, re-record or edit-with-plugin actions should trigger createChapterTake() given that the number of recorded items equals to the chunk total. Also, we defer the chapter compilation check until the SaveRecording action.

- Subscribe to the chapter-selected-take relay to correctly reflect the status of Narration Menu after resetting chapter.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/928)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/928)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @AnonymousWalker and @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 123 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/928#-NlOH0nM8cXAsfwX8ODr:-NlOH0nM8cXAsfwX8ODs:b-g00ri8) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/e1e221cf7179feb244bd5b828471392e400aceef/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L123)):*
> ```Kotlin
>     val snackBarObservable: PublishSubject<String> = PublishSubject.create()
> 
>     private val chapterCompleted: Boolean
> ```

the use of the term potentiallyFinished is intentional and I don't want it renamed.

This is because - we are not requiring the user to manually save (if we did, then we would know the chapter was indeed finished).

It's only potentially finished because a significant amount of work could continue to be done when audio exists for all recordable items. The user could in theory perform 90% of all work in this chapter AFTER this point. So we don't know if it's finished, it's only potentially finished.
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 127 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/928#-NlOHw7s2wmnJwdhqTFK:-NlOHw7s2wmnJwdhqTFL:b-t2dvmd) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/e1e221cf7179feb244bd5b828471392e400aceef/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L127)):*
> ```Kotlin
> 
>     private val disposables = CompositeDisposable()
>     private val chapterDisposable = CompositeDisposable()
> ```

why does this need to be a separate disposable?
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 330 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/928#-NlOIm03AvKMSLeX6wtF:-NlOIm03AvKMSLeX6wtG:b-t06a98) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/e1e221cf7179feb244bd5b828471392e400aceef/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L330)):*
> ```Kotlin
>     }
> 
>     private fun loadChapter(chapter: Chapter) {
> ```

why is this private and the other is public? This is perfectly fine to be used as an alternative to the other one?

PR #653, opened 2022-08-11
Test for daos

AnonymousWalker:
This change is [Reviewable](https://reviewable.io/reviews/bible-translation-tools/orature/653)

PR #938, opened 2023-12-14
Chapter take modifier (OT-789)

darrellcolehill:
Prevents user home/chapter navigation while bouncing audio.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)


darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt` line 50 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldmvRfE9baOHR6p98r:-NldmvRfE9baOHR6p98s:bj6j3qs) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt#L50)):*
> ```Kotlin
>  * provided by the markers parameter
>  */
> fun bounceAudio(bouncedAudio: File, reader: AudioFileReader, markers: List<AudioMarker>) {
> ```

Moved bounceAudio to AudioFileUtils because it seems to make more sense here and the functionality was needed outside of the Narration class. 
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterTakeAudioBouncer.kt` line 23 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldnIFg7wPmCC1P3NzJ:-NldnIFh7tngUMnT14-S:b-pi94rg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterTakeAudioBouncer.kt#L23)):*
> ```Kotlin
>     init {
>         val thread = Thread(this)
>         thread.start()
> ```

Starting up a thread in the init seems very convenient, however, I worry that it may cause leaks or issues trying to stop the thread later on
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterTakeAudioBouncer.kt` line 32 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldoDUQEQOZPrCeQOU9:-NldoDUQEQOZPrCeQOUA:bj2nd4u) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterTakeAudioBouncer.kt#L32)):*
> ```Kotlin
>             val currentTime = System.currentTimeMillis()
>             try {
>                 if (queue.size > 0 && currentTime - lastBounceTime > bouncingTimeThresholdInMilliseconds) {
> ```

Only bounces the audio if the queue size is > 0 and 2 seconds have passed since the last time audio has finished bouncing. This helps reduce the amount of times we are bouncing the audio when the user spams undo/redo (or any other action that requires us to bounce the audio)
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterTakeAudioBouncer.kt` line 61 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldnXyo-HHaO0pLOAbS:-NldnXyo-HHaO0pLOAbT:b-ithvng) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/ChapterTakeAudioBouncer.kt#L61)):*
> ```Kotlin
> 
>     fun bounce(file: File, reader: AudioFileReader, activeVerses: List<AudioMarker>) {
>         lock.lock()
> ```

locks are to prevent the user from adding more items in the list while the following block of code is being executed in the run method. I am not a huge fan of having to introduce locks here, but I couldn't think of a more clever solution. 



_Code snippet:_
```Kotlin
                    // Removes irrelevant requests.
                    for (i in 0 until queue.size - 1 step 1) {
                        queue.take()
                    }
                    mostRecentBounceState = queue.take()
```
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 110 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nldoo8ICZavz-9LU9f1:-Nldoo8ICZavz-9LU9f2:bwt2x78) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L110)):*
> ```Kotlin
>         )
>         loadChapterIntoPlayer()
>         takeAudioModifier = if (chapter.getSelectedTake() != null) {
> ```

Not a huge fan of how I am assigning a value to the takeAudioModifier here. The idea is that I need to update the value of takeAudioModifier whenever a user navigates to a chapter. Either that chapter has a take and we can assign a value to the takeAudioModifier, or it does not have a take and we assign null. 
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 435 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldpUN-1A2qoXwfNKmU:-NldpUN-1A2qoXwfNKmV:bjk972t) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L435)):*
> ```Kotlin
>                     activeVerses
>                 )
>                 takeAudioModifier = NarrationTakeAudioModifier(take)
> ```

Assigns a value to the takeAudioModifier whenever a new take is created. 
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 124 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nldphqr1bxrM-pEvJe0:-Nldphqr1bxrM-pEvJe1:b-fj0gqt) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L124)):*
> ```Kotlin
>         .and(isRecordingProperty.not())
>         .and(isRecordingAgainProperty.not())
>         .and(chapterTakeProperty.isNull)
> ```

Prevents chapter takes from being made multiple times. I feel like this will need to be updated to account for users recording an entire chapter, then clicking undo some number of times until they no longer have all verses recorded.

mXaln (changes requested):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: all files reviewed, 9 unresolved discussions (waiting on @AnonymousWalker, @darrellcolehill, and @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt` line 50 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldmvRfE9baOHR6p98r:-NlwJsxs78_Faw6nyDgn:bs9gze8) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt#L50)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

Moved bounceAudio to AudioFileUtils because it seems to make more sense here and the functionality was needed outside of the Narration class.
</blockquote></details>

Why is this function not inside AudioFileUtils class?
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt` line 11 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NlwLi5h-iiNvNnCzwIO:-NlwLi5h-iiNvNnCzwIP:bz82jb8) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt#L11)):*
> ```Kotlin
> import org.wycliffeassociates.otter.common.domain.audio.OratureAudioFile
> 
> class NarrationTakeAudioModifier(val take: Take, createNewAudioFile: Boolean = false) {
> ```

I didn't see where `createNewAudioFile` is true
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt` line 33 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NlwjU6m7sF2NmiRQh_-:-NlwjU6m7sF2NmiRQh_0:b-vr0hhx) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt#L33)):*
> ```Kotlin
>         clearNarrationMarkers()
>         addNarrationMarkers(markers)
>         audioFile.update()
> ```

I think this should also run in background thread. With big files it takes a lot of time and blocks undo/redo buttons.

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: all files reviewed, 9 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt` line 50 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldmvRfE9baOHR6p98r:-NleDyDn-rvPvmGMZcK9:b-4740hv) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt#L50)):*

Why is this function not inside AudioFileUtils class?
</blockquote></details>

The idea was to be able to allow us to be able to use just that method without having to instantiate an AudioFileUtils object, since it does need an IDirectoryProvider.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt` line 11 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NlwLi5h-iiNvNnCzwIO:-Nlx40v139f5mtYhRwoy:b4bqmbc) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt#L11)):*

I didn't see where `createNewAudioFile` is true
</blockquote></details>

That's a good point. When I started testing, I was thinking of using it when there is no existing wav file. That was helpful when writing test, but I think I was trying to solve a problem that we do not actually have in production, since the NarrationTakeAudioModifier only exists when a take is created.

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: all files reviewed, 9 unresolved discussions (waiting on @AnonymousWalker, @darrellcolehill, and @jsarabia)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt` line 50 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldmvRfE9baOHR6p98r:-NlxFqw97KHU0A2uf9_7:b-d8f30y) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt#L50)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

The idea was to be able to allow us to be able to use just that method without having to instantiate an AudioFileUtils object, since it does need an IDirectoryProvider.
</blockquote></details>

Since it's injected in `Narration` class, you could pass it to your `NarrationTakeAudioModifier` class in constructor.

darrellcolehill (commented):
UPDATE: there is no longer the 2 second delay between processing requests. Now, when a new bounce request is received, we interrupt the old one so we can begin work on the new one. 

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-:-Nm1Cd4G4qEqMTH3nP8M:b-hsh3at)* status: 1 of 9 files reviewed, 11 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 210 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nm1D-Vw9ANB-aJIh325:-Nm1D-Vw9ANB-aJIh326:b-nuncev) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bcd377b7fe682d728dd41145739d17601ac0838a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L210)):*
> ```Kotlin
>             // TODO: Possibly show the user a progress bar and message stating that the changes need to be saved
>             //  before navigating to a new chapter
>             logger.error("Undocking NarrationViewModel before chapter take modification is complete")
> ```

Right now, the chapter navigation buttons are disabled, however, this is here for if the user tries to leave the page by pressing the home button. My understanding of the ViewModel is that the undock is typically called when leaving the View that the ViewModel is associated with. 
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 335 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nm1Dlsh2qankD_ZnVf0:-Nm1Dlsh2qankD_ZnVf1:b8vltn7) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bcd377b7fe682d728dd41145739d17601ac0838a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L335)):*
> ```Kotlin
>             // TODO: Possibly show the user a progress bar and message stating that the changes need to be saved
>             //  before navigating to a new chapter
>             logger.error("TRYING TO NAVIGATE CHAPTER BEFORE TAKE MODIFICATION IS COMPLETE")
> ```

This is not currently necessary because I am disabling the chapter navigation buttons while bouncing the audio, however, since I am not sure if we want to always disable the buttons with no message so I left this here.

AnonymousWalker:
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 110 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nldoo8ICZavz-9LU9f1:-NmBv2_h34hGBy4L0BRo:by5aq9s) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L110)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

Not a huge fan of how I am assigning a value to the takeAudioModifier here. The idea is that I need to update the value of takeAudioModifier whenever a user navigates to a chapter. Either that chapter has a take and we can assign a value to the takeAudioModifier, or it does not have a take and we assign null.
</blockquote></details>

= chapter.getSelectedTake()?.let { NarrationTakeAudioModifier(it) }

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 1 of 9 files reviewed, 11 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt` line 11 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NlwLi5h-iiNvNnCzwIO:-NmBv_VG6GzDjw9uhdpZ:b-r4gvob) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt#L11)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

That's a good point. When I started testing, I was thinking of using it when there is no existing wav file. That was helpful when writing test, but I think I was trying to solve a problem that we do not actually have in production, since the NarrationTakeAudioModifier only exists when a take is created.
</blockquote></details>

can we always create a take file if `!file.exists()`? I would try to avoid having a boolean in the class constructor, even in function parameter, if possible.
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 335 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nm1Dlsh2qankD_ZnVf0:-NmBwbuMAcY4TvT7udut:b-n2wba) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bcd377b7fe682d728dd41145739d17601ac0838a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L335)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

This is not currently necessary because I am disabling the chapter navigation buttons while bouncing the audio, however, since I am not sure if we want to always disable the buttons with no message so I left this here.
</blockquote></details>

can we use normal-case instead of all CAPITALs?

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 1 of 9 files reviewed, 11 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 110 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nldoo8ICZavz-9LU9f1:-NmfyIW69EghxfvYloro:b-aja6cc) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L110)):*

= chapter.getSelectedTake()?.let { NarrationTakeAudioModifier(it) }
</blockquote></details>

I like that way more. 
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 335 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nm1Dlsh2qankD_ZnVf0:-NmflLnM6t9pyrjsr8eq:b-r0eth8) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/bcd377b7fe682d728dd41145739d17601ac0838a/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L335)):*

can we use normal-case instead of all CAPITALs?
</blockquote></details>

Absolutely! Didn't mean to let this get through

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 1 of 9 files reviewed, 12 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationAudioBouncerTaskRunner.kt` line 35 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nmm1daGB1_IRkAHEGjB:-Nmm1daGB1_IRkAHEGjC:b-g7wigs) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a6fae5336d13743650e76846ed9c887cdb0ac3f7/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationAudioBouncerTaskRunner.kt#L35)):*
> ```Kotlin
>             }
>             .doOnComplete {
>                 logger.info("Successfully completed!")
> ```

may need a name of what completed here

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 1 of 9 files reviewed, 14 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 125 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nmm3VpV7QrIjvXIZz-m:-Nmm3VpV7QrIjvXIZz-n:b8kw3gt) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a6fae5336d13743650e76846ed9c887cdb0ac3f7/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L125)):*
> ```Kotlin
>         )
> 
>         // Verify that we have the expected amount of audio data
> ```

here are some references to better ways of describing test objectives. I applied them to a lot of places where tests are complicated.

https://github.com/Bible-Translation-Tools/Orature/pull/653#discussion_r982849763
https://github.com/Bible-Translation-Tools/Orature/pull/653#discussion_r982859789
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 133 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nmm4OOx1dQK_i9901o0:-Nmm4OOx1dQK_i9901o1:bw549ey) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a6fae5336d13743650e76846ed9c887cdb0ac3f7/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L133)):*
> ```Kotlin
>         }
> 
>         val newAudioMarkers = listOf<AudioMarker>(
> ```

repeated object construction like this could be refactored. Tests are also functions and they should not be long.

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt` line 11 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NlwLi5h-iiNvNnCzwIO:-NmfyajW8TXEjA_k5Cvq:b11osty) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt#L11)):*

can we always create a take file if `!file.exists()`? I would try to avoid having a boolean in the class constructor, even in function parameter, if possible.
</blockquote></details>

I like that much better! 
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 125 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nmm3VpV7QrIjvXIZz-m:-NnA295l4FySJV0roGYT:bq31ayd) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a6fae5336d13743650e76846ed9c887cdb0ac3f7/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L125)):*

here are some references to better ways of describing test objectives. I applied them to a lot of places where tests are complicated.

https://github.com/Bible-Translation-Tools/Orature/pull/653#discussion_r982849763
https://github.com/Bible-Translation-Tools/Orature/pull/653#discussion_r982859789
</blockquote></details>

Thank you! I have been having a hard time figuring out how to describe the test without having a monster of a function name. 
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 133 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nmm4OOx1dQK_i9901o0:-NnA2t5oCv13xtAOWEPR:bdvyd03) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a6fae5336d13743650e76846ed9c887cdb0ac3f7/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L133)):*

repeated object construction like this could be refactored. Tests are also functions and they should not be long.
</blockquote></details>

Agreed. The hard-coding was getting out of hand. 
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 163 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnA2gnHADrkirgl_UuF:-NnA2gnHADrkirgl_UuG:b-7cb7yj) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L163)):*
> ```Kotlin
> 
> 
>     // Test that Wav file metadata is updated properly when keeping the same markers, but changing their location
> ```

@AnonymousWalker wasn't sure if I should make this more/less detailed

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 9 files reviewed, 13 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 163 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnA2gnHADrkirgl_UuF:-NnB6txJ86Tj-LhWw6sV:b-2evw9s) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L163)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

@AnonymousWalker wasn't sure if I should make this more/less detailed
</blockquote></details>

this sounds like `testMoveMarkers`
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 201 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnB8bOXEyoN97kO9g43:-NnB8bOXEyoN97kO9g44:b-sijrta) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L201)):*
> ```Kotlin
>     // Test that Wav file metadata is updated properly when using a different amount of markers with different locations
>     @Test
>     fun `modifyMetadata with different amount of markers, same labels, and with different locations`() {
> ```

`testMoveThenDeleteMarkers`?
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 237 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnB7S3s24GPRNLbVc_U:-NnB7S3s24GPRNLbVc_V:b-ic8n5r) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L237)):*
> ```Kotlin
> 
>     @Test
>     fun `modifyMetadata multiple times with same markers`() {
> ```

if we want to test the consistency of `modifyMetadata()` outcome, we can consider running a loop of 3-5 executions at least

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 9 files reviewed, 13 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 163 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnA2gnHADrkirgl_UuF:-NnEx5dGArz-ABEjza0Q:bkxba6r) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L163)):*

this sounds like `testMoveMarkers`
</blockquote></details>

Thanks for that clarification! I like swapping out the verbose function name for a more straight forward function name with a comment to give clarification. 
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 201 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnB8bOXEyoN97kO9g43:-NnF4im6BTPjuSZ0r1tp:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L201)):*

`testMoveThenDeleteMarkers`?
</blockquote></details>

Sounds good
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 237 at r6](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnB7S3s24GPRNLbVc_U:-NnF4vRWBiUnBOYPt8ws:bqog80t) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3887f0b2c535e9b49cf84fea11a4c76f955cffb2/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L237)):*

if we want to test the consistency of `modifyMetadata()` outcome, we can consider running a loop of 3-5 executions at least
</blockquote></details>

Agreed

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 10 files reviewed, 13 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt` line 33 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NlwjU6m7sF2NmiRQh_-:-NnixFyb2eAcz2cAjoNw:bqog80t) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeAudioModifier.kt#L33)):*

I think this should also run in background thread. With big files it takes a lot of time and blocks undo/redo buttons.
</blockquote></details>

Agreed
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt` line 50 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NldmvRfE9baOHR6p98r:-NnixRAW2Y-3JbunWD3q:b-hcdr7r) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/AudioFileUtils.kt#L50)):*

Since it's injected in `Narration` class, you could pass it to your `NarrationTakeAudioModifier` class in constructor.
</blockquote></details>

Thanks for pointing that out!

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 10 files reviewed, 16 unresolved discussions (waiting on @AnonymousWalker, @darrellcolehill, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioBouncer.kt` line 38 at r8](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnnVhFb9Ygf2QJ1MZqf:-NnnVhFb9Ygf2QJ1MZqg:b-76r39z) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3d246ac827f9061cf5f06edc278aead1a4ce92ec/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/audio/AudioBouncer.kt#L38)):*
> ```Kotlin
>         }
> 
>         val wav = WavFile(bouncedAudio, 1, 44100, 16)
> ```

there are probably constants already defined for these which you should use

Also, we could think about doing this more generically. Here you have hard coded 44.1khz 16 bit mono. What if we wanted to support something different? What should define this? What is the configuration of the input? What would happen if the input is of a different configuration than the output you're hardcoding here?
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 110 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nldoo8ICZavz-9LU9f1:-NnnYWkP0giFPddpte5L:b-fo8kzj) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L110)):*
<details><summary><i>Previously, darrellcolehill (Darrell Cole Hill) wrote…</i></summary><blockquote>

I like that way more.
</blockquote></details>

I think I would rather you store a reference to the selected take, not a class called NarrationTakeAudioModifier

NarrationTakeAudioModifier has no need to be a class, it can be refactored to being a function.
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 110 at r8](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnnXkY53OKdx1Jc2nDE:-NnnXkY53OKdx1Jc2nDF:b-w1u0va) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3d246ac827f9061cf5f06edc278aead1a4ce92ec/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L110)):*
> ```Kotlin
>         var verseCount = 1
>         for (i in 0 until numberOfMarkers) {
>             if (i == 0) {
> ```

let's replace this if/elseif/else with a when statement
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 132 at r8](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnnXrM45E8hkjTfYvfV:-NnnXrM45E8hkjTfYvfW:b4uve55) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3d246ac827f9061cf5f06edc278aead1a4ce92ec/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L132)):*
> ```Kotlin
>             if (it == markerToMove) {
>                 val newLocation = it.location + delta
>                 if (it::class == BookMarker::class) {
> ```

same as above, use a when statement

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 20 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt` line 110 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nldoo8ICZavz-9LU9f1:-NolwbqL5qBz1nNWWTu7:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a723ed0fd75040a4bc3b4f0980bdcbddd6a98c77/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/Narration.kt#L110)):*

I think I would rather you store a reference to the selected take, not a class called NarrationTakeAudioModifier

NarrationTakeAudioModifier has no need to be a class, it can be refactored to being a function.
</blockquote></details>

Sounds good
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 47 at r11](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NolxDJvA57P-LHv8tNj:-NolxDJvA57P-LHv8tNk:bll3fsi) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/cf51bbcb7f92afdef961451a9de9ec3157c08a4b/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L47)):*
> ```Kotlin
>             busyStatusEmitter = it
>         }.publish()
>     private var isBouncingAudio = false
> ```

In my gut, I feel that this should be atomic, but I couldn't think of a case where it would be necessary
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/NavigationMediator.kt` line 100 at r11](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NolzRRH8foYqJsVsGU2:-NolzRRH8foYqJsVsGU3:b-x55oid) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/cf51bbcb7f92afdef961451a9de9ec3157c08a4b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/NavigationMediator.kt#L100)):*
> ```Kotlin
>         subscribe<NavigationRequestEvent> {
> 
>             if (blockNavigationEvents.value) {
> ```

Took some inspiration from how the shouldBlockWindowCloseRequest Boolean is used in the OtterApp class.

Fires a NavigationRequestBlockedEvent with the event passed as a parameter so it can be fired again later on. 
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 357 at r11](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nom0jDfA9GB4qDbM6fd:-Nom0jDfA9GB4qDbM6fe:b5vk1tm) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/cf51bbcb7f92afdef961451a9de9ec3157c08a4b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L357)):*
> ```Kotlin
> 
>         // Prevents navigating to chapter while modifying chapter take audio
>         if (isModifyingTakeAudioProperty.value) {
> ```

It could be beneficial to wrap this function (in something like tryToNavigateChapter) and put the if-statement in the wrapper. The main reason I didn't do that is because I couldn't think of a better name for the wrapper method. 
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 816 at r11](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nom0EGq0bG5iYevZFM9:-Nom0EGq0bG5iYevZFMA:b-r4p24e) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/cf51bbcb7f92afdef961451a9de9ec3157c08a4b/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L816)):*
> ```Kotlin
>                         openSavingModalProperty.set(false)
> 
>                         if (pendingChapterToLoadProperty.value != null) {
> ```

Not crazy about this if-elseif structure because it seems that it could blow up if more events are added to the list of things to block while modifying take audio.

AnonymousWalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 24 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 29 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NormObd3szRw__QN_ca:-NormObd3szRw__QN_cb:b8t3mjq) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L29)):*
> ```Kotlin
> 
> 
> object NarrationTakeModifierTaskRunner {
> ```

I think we can remove the word "TaskRunner" and still communicate the same meaning of this object
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 43 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NorkqB01tHPGyjyi80d:-NorkqB01tHPGyjyi80e:b-dyvvwg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L43)):*
> ```Kotlin
> 
>     private var busyStatusEmitter: ObservableEmitter<TaskRunnerStatus>? = null
>     val busyStatus: ConnectableObservable<TaskRunnerStatus> = Observable
> ```

`runnerStatus` or `status` maybe a better name for this observable
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 135 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Norl97a6_4PZslII_yd:-Norl97a6_4PZslII_ye:bxnp0ia) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L135)):*
> ```Kotlin
>             busyStatus
>                 .takeWhile { busy ->
>                     busy == TaskRunnerStatus.UPDATE_MARKERS_WAITING || busy == TaskRunnerStatus.BOUNCING_AUDIO
> ```

`busy -> ` sounds like a boolean. How about `status -> `?
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 142 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Norla2y3PmsJZPpeTYl:-Norla2y3PmsJZPpeTYm:bdjd4zq) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L142)):*
> ```Kotlin
>                 }
>                 .doOnDispose {
>                     logger.info("Disposing unnecessary marker update")
> ```

"Cancelling stale update markers task"?

AnonymousWalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 26 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 24 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NornFOvDbxjihdpWyau:-NornFOvDbxjihdpWyav:b-4g3oxu) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L24)):*
> ```Kotlin
>     BOUNCING_AUDIO,
>     UPDATING_MARKERS,
>     UPDATE_MARKERS_WAITING,
> ```

it's quite challenging to understand this status by looking at its name, at least for me
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 121 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Normyht4VmshP7Nx99s:-Normyht4VmshP7Nx99t:b293116) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L121)):*
> ```Kotlin
>     fun updateMarkers(file: File, markers: List<AudioMarker>) {
> 
>         if (!isBouncingAudio) {
> ```

It would be less confusing to swap the condition here to 
`if (isBouncingAudio) { } else { }`

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 26 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 132 at r8](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnnXrM45E8hkjTfYvfV:-NornllMDLjED076JzkQ:b-3vyuml) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3d246ac827f9061cf5f06edc278aead1a4ce92ec/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L132)):*

same as above, use a when statement
</blockquote></details>

```
when(it) { 
is BookMarker -> ...
is ChapterMarker -> ...
else -> ...
}
```

AnonymousWalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 27 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/dialog/SavingModal.kt` line 8 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NoroCZM-F1oh47tb0Ig:-NoroCZM-F1oh47tb0Ih:b-hzybt0) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/dialog/SavingModal.kt#L8)):*
> ```Kotlin
> import tornadofx.*
> 
> class SavingModal : OtterDialog() {
> ```

can we use the existing `LoadingModal` instead? I've refactored it so that we could set the message dynamically

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 26 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 29 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NormObd3szRw__QN_ca:-Nos-hEA-pKnUluzK_Ud:bqog80t) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L29)):*

I think we can remove the word "TaskRunner" and still communicate the same meaning of this object
</blockquote></details>

Agreed
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 43 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NorkqB01tHPGyjyi80d:-NorqP3u7yn1Cf_1nnCi:bqog80t) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L43)):*

`runnerStatus` or `status` maybe a better name for this observable
</blockquote></details>

Agreed
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 121 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Normyht4VmshP7Nx99s:-NorqnQk9IFQ_6t0cv__:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L121)):*

It would be less confusing to swap the condition here to 
`if (isBouncingAudio) { } else { }`
</blockquote></details>

Sounds good
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 135 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Norl97a6_4PZslII_yd:-NorrPYn4x7oysGQWUp2:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L135)):*

`busy -> ` sounds like a boolean. How about `status -> `?
</blockquote></details>

Sounds good
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 142 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Norla2y3PmsJZPpeTYl:-Norrj_U4q1EC1IzicTi:b-rvkbqg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L142)):*

"Cancelling stale update markers task"?
</blockquote></details>

I like that much more. It will be much easier to decipher when we get logs back from the field. 
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 110 at r8](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnnXkY53OKdx1Jc2nDE:-Nors9b8ByYQSz2qKP9q:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3d246ac827f9061cf5f06edc278aead1a4ce92ec/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L110)):*

let's replace this if/elseif/else with a when statement
</blockquote></details>

Sounds good
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 132 at r8](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NnnXrM45E8hkjTfYvfV:-NorsrvD0Tlz2uKw1crF:b-yqdoz6) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/3d246ac827f9061cf5f06edc278aead1a4ce92ec/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L132)):*

```
when(it) { 
is BookMarker -> ...
is ChapterMarker -> ...
else -> ...
}
```
</blockquote></details>

That is so much nicer! Thanks for pointing this out! 
___
*[`jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/dialog/SavingModal.kt` line 8 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NoroCZM-F1oh47tb0Ig:-Nort1P97_NOZ6XKUK_l:b-s4pk3w) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/dialog/SavingModal.kt#L8)):*

can we use the existing `LoadingModal` instead? I've refactored it so that we could set the message dynamically
</blockquote></details>

Absolutely! The only hesitation I have is if we want to have functionality for click-away events, but we should be able to add that in no problem here if we needed to.

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 26 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt` line 24 at r12](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NornFOvDbxjihdpWyau:-NosF2SA-5tZqFRgLtUY:b-oqp47r) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/d3e8b09117e657d7fc8aff3d1eeb67684f1e3e15/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTaskRunner.kt#L24)):*

it's quite challenging to understand this status by looking at its name, at least for me
</blockquote></details>

Thanks again for explaining why the waiting state was unnecessary!

AnonymousWalker (approved):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 15 unresolved discussions (waiting on @darrellcolehill, @jsarabia, and @mXaln)

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 9 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationAudioBouncerTaskRunner.kt` line 35 at r5](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-Nmm1daGB1_IRkAHEGjB:-Novyqsj4UU3VKw5Z_Jc:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/a6fae5336d13743650e76846ed9c887cdb0ac3f7/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationAudioBouncerTaskRunner.kt#L35)):*

may need a name of what completed here
</blockquote></details>

Done.

jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 10 unresolved discussions (waiting on @AnonymousWalker, @darrellcolehill, and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 81 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowN98U5GAuRfBKuppf:-NowN98U5GAuRfBKuppg:b3002it) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L81)):*
> ```Kotlin
>             val outputStream = FileOutputStream(file, false)
> 
>             val byteArray = ByteArray(numberOfBytes) { 1 }
> ```

Why are we writing ones?
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 82 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowNK3w-B1k2UJ7InjT:-NowNK3w-B1k2UJ7InjU:b6ppuzn) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L82)):*
> ```Kotlin
> 
>             val byteArray = ByteArray(numberOfBytes) { 1 }
>             outputStream.write(byteArray)
> ```

use this in a .use block so that it closes safely
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 84 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowNORL4iu3YNzjuxcB:-NowNORM7D2ZQIJK7O3n:b-roapbg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L84)):*
> ```Kotlin
>             outputStream.write(byteArray)
> 
>             outputStream.close()
> ```

if the write throws, this won't happen, it should be put in a finally, but that's also not quite good enough, but a use block handles this correctly for you
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 347 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowKRYF5G1ZC0XM4Daw:-NowKRYF5G1ZC0XM4Dax:bdawaz9) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L347)):*
> ```Kotlin
> 
>     // Prevents navigating to chapter while modifying chapter take audio
>     fun tryToNavigateChapter(chapterNumber: Int) {
> ```

We aren't "trying" anything here- we're doing it.
We WILL navigate, but it may just have to get queued up to wait on existing work.

the word try implies we will ATTEMPT to navigate, but that may not actually happen, in which there is a possibility that nothing happens.

As such, we should rename this to not have the method name be misleading.
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifier.kt` line 167 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowEr2NAMbWZABm1mid:-NowEr2NAMbWZABm1mie:b-pwaqps) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifier.kt#L167)):*
> ```Kotlin
>                     currentMarkerUpdateTask?.dispose()
>                 }
>                 .subscribe().let {
> ```

.let should go on a new line like every other function being chained here

darrellcolehill (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: 0 of 15 files reviewed, 10 unresolved discussions (waiting on @AnonymousWalker, @jsarabia, and @mXaln)

___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 81 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowN98U5GAuRfBKuppf:-NowTweM8_CDvIokMK9m:b-p24vjv) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L81)):*

Why are we writing ones?
</blockquote></details>

No particular reason. The reason for these test is mostly to verify that we are not experiencing loss of data when moving markers while bouncing. So I was not worried about actual content of the audio file, but rather, the size. 

I could improve it by adding an additional parameter called something like "valueToWrite" so we can control what you're writing to the file at specific times. I could see that being useful for more flushed out test in the future. 
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 82 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowNK3w-B1k2UJ7InjT:-NowS1v1EmVgHoto5W1E:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L82)):*

use this in a .use block so that it closes safely
</blockquote></details>

Sounds good
___
*[`common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt` line 84 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowNORL4iu3YNzjuxcB:-NowRyRCBn8Ly-9dgCVE:b45dymg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/test/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifierTest.kt#L84)):*

if the write throws, this won't happen, it should be put in a finally, but that's also not quite good enough, but a use block handles this correctly for you
</blockquote></details>

Thank you for reminding me! I forget that .use handles that for me. 
___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt` line 347 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowKRYF5G1ZC0XM4Daw:-NowVpLr3k62h3aaxTCQ:b-8e5dqk) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/narration/NarrationViewModel.kt#L347)):*

We aren't "trying" anything here- we're doing it.
We WILL navigate, but it may just have to get queued up to wait on existing work.

the word try implies we will ATTEMPT to navigate, but that may not actually happen, in which there is a possibility that nothing happens.

As such, we should rename this to not have the method name be misleading.
</blockquote></details>

That makes a lot of sense. I was having a hard time thinking of another way to phrase it without leaking implementation details to the UI code. 
___
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifier.kt` line 167 at r16](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-NowEr2NAMbWZABm1mid:-NowRrMK7m34pbCSVQtO:b-hfghm4) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/45a467b7fd067698ccf293ebf49c77b02789f6b9/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/NarrationTakeModifier.kt#L167)):*

.let should go on a new line like every other function being chained here
</blockquote></details>

Sounds good

jsarabia (commented):
> Reviewed 2 of 8 files at r2, 1 of 5 files at r8, 1 of 4 files at r10, 3 of 10 files at r11, 1 of 2 files at r13, 3 of 7 files at r14, 4 of 4 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938)* status: all files reviewed, 10 unresolved discussions (waiting on @AnonymousWalker and @mXaln)

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/938#-:-NowtZOb0eEfpOAiuoR9:bnfp4nl)* status: all files reviewed, 10 unresolved discussions (waiting on @AnonymousWalker and @mXaln)

PR #946, opened 2023-12-19
Enable narration chapters for export, even if they are not completed (OT-794)

mXaln:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/946)


AnonymousWalker (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/946)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ExportProjectViewModel.kt` line 71 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/946#-Nm2bNZq8pjeAJ_waIoC:-Nm2bNZq8pjeAJ_waIoD:bu1tves) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0ba560f237d634ded8ac84b9b5c9cd6c27150430/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ExportProjectViewModel.kt#L71)):*
> ```Kotlin
>                             // This narration progress does not reflect real progress
>                             // It's here just to enable this chapter in the export dialog
>                             hasInProgressNarration(workbook, chapter) -> 0.01
> ```

I would recommend fixing the underlying problem of incorrect progress value and not adding a "mask" on top of the bug. This only fixes the symptom.

The problem: narration doesn't work with individual Chunks like translation does, thus making it a little different when it comes to evaluating the progress of a chapter. There are two ways I could come up with:

1. Parse the serialized verse `.json` file to get the number of recorded items and calculate the ratio OR 
2. Use the `ChapterRepresentation` class to access the _totalVerses_ and filter the _placed_ verse nodes to calculate the ratio.

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/946)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ExportProjectViewModel.kt` line 71 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/946#-Nm2bNZq8pjeAJ_waIoC:-Nm6yZWm8sxX_mwQNskO:b-896fix) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/0ba560f237d634ded8ac84b9b5c9cd6c27150430/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ExportProjectViewModel.kt#L71)):*

I would recommend fixing the underlying problem of incorrect progress value and not adding a "mask" on top of the bug. This only fixes the symptom.

The problem: narration doesn't work with individual Chunks like translation does, thus making it a little different when it comes to evaluating the progress of a chapter. There are two ways I could come up with:

1. Parse the serialized verse `.json` file to get the number of recorded items and calculate the ratio OR 
2. Use the `ChapterRepresentation` class to access the _totalVerses_ and filter the _placed_ verse nodes to calculate the ratio.
</blockquote></details>

Done.

AnonymousWalker (approved):
@jsarabia do you have any further comment on the workaround? We get all chapters in Book using count() because the chapters observable is wrapped with `.cache()`

PR #958, opened 2024-01-03
Handle migration of translation projects - chunk from verses

AnonymousWalker:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/958)


jsarabia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/958)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChunkingViewModel.kt` line 227 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/958#-NnFo8cGEGmmxo3LMUNG:-NnFo8cGEGmmxo3LMUNH:bmx4t5j) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/6db9880f16654ef87e9d321697cd483220c6ad54/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChunkingViewModel.kt#L227)):*
> ```Kotlin
>      * the chapter contains derived chunks.
>      */
>     private fun getChunkMarkersFromAudio(audio: OratureAudioFile): List<ChunkMarkerModel> {
> ```

This feels pretty messy.

First, I feel like it's not appropriate for a ViewModel.

Second, I don't like the idea of having case checks every time we open this page to format various versions of states the content can be in. We're in Orature 3. When you get to this page, we're deep in the heart of Orature 3. All content by now needs to be up to date, otherwise we'll have to be perpetually paranoid- this is not much different than not handling null checks early on and propagating them deeper and deeper down a call stack. No.

If content is not Orature 3 content, it comes in from two places: Import, and File System. It needs to get migrated through the two isolated points of entry. Then, we can trust that if it was properly migrated there, then everywhere else in the code is able to just assume the state of a project is up to date.

AnonymousWalker (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/958)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChunkingViewModel.kt` line 227 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/958#-NnFo8cGEGmmxo3LMUNG:-NnFvSe-5oVC5o6vF70N:b-tbr5om) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/6db9880f16654ef87e9d321697cd483220c6ad54/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/viewmodel/ChunkingViewModel.kt#L227)):*

This feels pretty messy.

First, I feel like it's not appropriate for a ViewModel.

Second, I don't like the idea of having case checks every time we open this page to format various versions of states the content can be in. We're in Orature 3. When you get to this page, we're deep in the heart of Orature 3. All content by now needs to be up to date, otherwise we'll have to be perpetually paranoid- this is not much different than not handling null checks early on and propagating them deeper and deeper down a call stack. No.

If content is not Orature 3 content, it comes in from two places: Import, and File System. It needs to get migrated through the two isolated points of entry. Then, we can trust that if it was properly migrated there, then everywhere else in the code is able to just assume the state of a project is up to date.
</blockquote></details>

I agree this code could be refactored to a use case object - not located in the ViewModel.

Initially, I attempted to handle verse-by-verse migration during InitializeProjects - Import. However, projects in Orature 1 may not always have source audio with them to extract verse markers and `.cue` files for populating chunk markers. 

The purpose of this is to handle verse-by-verse fallback and it can also support Verse By Verse mode which will be added later in the design

Issue #967, opened 2024-01-07
Build fails for JDK 13+


mbr4477:
All tested on macOS Sonoma 14.2.1 (23C71) M3 (aarch64).

## JDK 13+ Issues
Thanks to a new function signature for `java.nio.file.FileSystems.newFileSystem` (see [JDK 13 release notes](https://www.oracle.com/java/technologies/javase/13-relnote-issues.html)), compilation fails for `jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/domain/resourcecontainer/project/ZipEntryTreeBuilder.kt` because the second argument `null` is now ambiguous.

### To Reproduce
Install JDK >= 13 and <= 20. Follow the readme instruction to install and run. The build will fail with an error like this:

```
e: <path>/Orature/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/domain/resourcecontainer/project/ZipEntryTreeBuilder.kt:38:28 Overload resolution ambiguity: 
public open fun newFileSystem(p0: Path!, p1: ClassLoader!): FileSystem! defined in java.nio.file.FileSystems
public open fun newFileSystem(p0: Path!, p1: (Mutable)Map<String!, *>!): FileSystem! defined in java.nio.file.FileSystems
```

## JDK 21 Issues
In addition, builds fail on JDK 21 thanks to [this issue](https://youtrack.jetbrains.com/issue/KT-60507/Kapt-IllegalAccessError-superclass-access-check-failed-using-java-21-toolchain) which may only be fixed in Kotlin 2.0.

### To Reproduce
Install JDK 21. Follow the readme instructions to install and run. The build will fail with an error like this:
```
e: java.lang.IllegalAccessError: superclass access check failed:
class org.jetbrains.kotlin.kapt3.base.javac.KaptJavaCompiler (in unnamed module @0x531ecf7)
cannot access class com.sun.tools.javac.main.JavaCompiler (in module jdk.compiler)
because module jdk.compiler does not export com.sun.tools.javac.main to unnamed module @0x531ecf7
```

jsarabia:
Fixed with #1074

PR #968, opened 2024-01-07
Fix function overload issue for JDK 13+ and update README with JDK max version allowed

mbr4477:
Fixes overload issue described in #967 using suggested fix from the [JDK 13 release notes](https://www.oracle.com/java/technologies/javase/13-relnote-issues.html) by casting null value with `null as java.lang.ClassLoader?` to resolve the ambiguity.

Since build fails on JDK 21.0.1, updates README to indicate max version of JDK is v20 because of [this issue](https://youtrack.jetbrains.com/issue/KT-60507/Kapt-IllegalAccessError-superclass-access-check-failed-using-java-21-toolchain). Probably worth putting in, but if installers package the right JRE it might be irrelevant unless building from source. 

> As a note, the macOS distributable from the installer list is for x86_64 and doesn't work with Apple silicon (arm64).
> Click-to-run after the install caused a crash, and running from the command line gives various versions of:
> ```
> Loading library ... from resource failed: ... 
>     (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
> ```

Built and tested with JDK 17.0.2 (all tests passing) and JDK 20.0.2 (all tests passing) on macOS Sonoma 14.2.1 (23C71) with M3 (aarch64).

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/968)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,11 @@ More information can be found [here.](https://bibletranslationtools.org/tool/ora
 </p>
 
 # Usage
-Orature runs on OpenJDK 11+. Installers for Windows, Debian-based Linux, and Mac are available in the Releases section of the repository on GitHub.
+Orature runs on OpenJDK >= 11 and <= 20. Installers for Windows, Debian-based Linux, and Mac are available in the Releases section of the repository on GitHub.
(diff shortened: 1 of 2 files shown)


mbr4477:
Failing CI due to missing secret key?

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/968#-:-Nnei5sOAqig7nWj-Nh7:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #1010, opened 2024-02-02
Add additional events when to clear waveform

mXaln:
Added `BeforeNavigationEvent` to track when to clear waveform in Translation mode. Attempts to clean waveform will happen when going to Home page, other chapter and other step.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010)

### DIFF
--- a/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeMarker.kt
+++ b/assets/src/main/kotlin/org/wycliffeassociates/otter/assets/initialization/InitializeMarker.kt
@@ -41,7 +41,7 @@ class InitializeMarker @Inject constructor(
 ) : Installable {
 
     override val name = "MARKER"
-    override val version = 32
+    override val version = 33
     private val log = LoggerFactory.getLogger(InitializeMarker::class.java)
 
     override fun exec(progressEmitter: ObservableEmitter<ProgressStatus>): Completable {
--- a/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/event/NavigationRequestEvent.kt
+++ b/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/event/NavigationRequestEvent.kt
@@ -24,6 +24,6 @@ import tornadofx.View
 class NavigationRequestEvent(val view: View) : FXEvent()
 class NavigationRequestBlockedEvent(val navigationRequest: NavigationRequestEvent) : FXEvent()
 class NavigateChapterEvent(val chapterNumber: Int) : FXEvent()
-class BeforeNavigationEvent : FXEvent()
+class TranslationNavigationEvent : FXEvent()
 
 object AppCloseRequestEvent : FXEvent()
--- a/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/event/NavigationRequestEvent.kt
+++ b/jvm/controls/src/main/kotlin/org/wycliffeassociates/otter/jvm/controls/event/NavigationRequestEvent.kt
@@ -24,5 +24,6 @@ import tornadofx.View
 class NavigationRequestEvent(val view: View) : FXEvent()
 class NavigationRequestBlockedEvent(val navigationRequest: NavigationRequestEvent) : FXEvent()
 class NavigateChapterEvent(val chapterNumber: Int) : FXEvent()
+class BeforeNavigationEvent : FXEvent()
 
 object AppCloseRequestEvent : FXEvent()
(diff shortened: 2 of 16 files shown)


AnonymousWalker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @mXaln)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt` line 58 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010#-Npe-ktl2EJy-hZl6Ywz:-Npe-ktl2EJy-hZl6Yx-:b-rkysqg) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/855a200659f031e8355b777cf8e97268dc848bee/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt#L58)):*
> ```Kotlin
>             val homePage = find<HomePage2>()
>             fire(NavigationRequestEvent(homePage))
>         }
> ```

`NavigationRequestEvent` was created to handle clean-up before navigating. You can use it instead of creating a new event class

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt` line 58 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010#-Npe-ktl2EJy-hZl6Ywz:-Npe6Nro66T3k7uvokSQ:b-txepsk) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/855a200659f031e8355b777cf8e97268dc848bee/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt#L58)):*

`NavigationRequestEvent` was created to handle clean-up before navigating. You can use it instead of creating a new event class
</blockquote></details>

Unfortunately I can't use `NavigationRequestEvent` in my case. At first, the views in Translation mode undock immediately when that event is fired, thus subscription to this event will be closed. Secondly, I'm using `BeforeNavigationEvent` not only when docking other views, but also when switching between chapters and steps.

jsarabia (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt` line 58 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010#-Npe-ktl2EJy-hZl6Ywz:-NpfJS1L5pAxTU0wbYXU:bve44ip) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/855a200659f031e8355b777cf8e97268dc848bee/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt#L58)):*

Unfortunately I can't use `NavigationRequestEvent` in my case. At first, the views in Translation mode undock immediately when that event is fired, thus subscription to this event will be closed. Secondly, I'm using `BeforeNavigationEvent` not only when docking other views, but also when switching between chapters and steps.
</blockquote></details>

Could we add a payload to the event for pre-navigation, which would in this case then carry the lambda to clean the images? So as to use the same NavigationEvent, but detach the execution of that method from the View which is docking status dependent?

AnonymousWalker (approved):

mXaln (commented):
> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010)* status: 0 of 16 files reviewed, 1 unresolved discussion (waiting on @AnonymousWalker and @jsarabia)

___
*[`jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt` line 58 at r1](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1010#-Npe-ktl2EJy-hZl6Ywz:-Nptz0yc8y-CcYdcpjA8:bdr1vsn) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/855a200659f031e8355b777cf8e97268dc848bee/jvm/workbookapp/src/main/kotlin/org/wycliffeassociates/otter/jvm/workbookapp/ui/components/AppBar.kt#L58)):*

Could we add a payload to the event for pre-navigation, which would in this case then carry the lambda to clean the images? So as to use the same NavigationEvent, but detach the execution of that method from the View which is docking status dependent?
</blockquote></details>

@jsarabia, I'm not sure how to achieve this

PR #1065, opened 2024-02-28
Handle unrecorded titles when migrating narration projects

AnonymousWalker:
Uses the index of verse based on the total list, not the recorded list, across the view models. Also, unrecorded titles from migrated projects require "prepending" mode rendering which is the same as record again.

This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1065)


AnonymousWalker:
*[`common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/teleprompter/TeleprompterStateTransition.kt` line 186 at r2](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1065#-NsA0SvP9r5g4SL28oyy:-NsA0SvP9r5g4SL28oyz:b-vjn64c) ([raw file](https://github.com/Bible-Translation-Tools/Orature/blob/555faba67ce99c7b85796f506089cb7e46ecdd3e/common/src/main/kotlin/org/wycliffeassociates/otter/common/domain/narration/teleprompter/TeleprompterStateTransition.kt#L186)):*
> ```Kotlin
>                     contexts[i].changeState(TeleprompterItemState.RECORD_AGAIN)
>                 } else if (contexts[i].state.type == TeleprompterItemState.RECORD_DISABLED) {
>                     contexts[i].changeState(TeleprompterItemState.RECORD)
> ```

It's unclear to me regarding the original intention of this `else` block, but it accidentally allowed both book and chapter titles to be in the "RECORD" state at the same time.

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1065#-:-Ntzy8Cz6CeV-v6h62b-:bnfp4nl)* status: 0 of 31 files reviewed, all discussions resolved

AnonymousWalker:
There are several major changes to how we do state machine transition in this PR. Because verses from Ot1 does not contain titles and we want to allow recording them, this introduces the "new" state of some verses recorded in the middle of the item list. In other words, we now support discontinuous recorded items.

Below are a few examples:

- "next verse" action will not assume the target will be at [index + 1]. Instead, we use the current index to find where the next item is based on its state. This is the case when only verse 1, 2 were recorded and the rest was not. Thus, the titles and verse 3, 4, 5... will be unrecorded. Going "next" from chapter title item, for example, should jump to verse 3 because the verse at [index + 1] has been already recorded (verse 1)
- updating the states of the remaining items when the current state changes requires taking into account the states of the discontiuous items recorded, too.
- we now use the index with respect to `totalVerses` instead of `recordedVerse` because the two lists are no longer starting at the same index, given that we have Ot1 projects not containing titles.

PR #1101, opened 2024-04-03
Fix crash deleting projects [OT-859]

AnonymousWalker:
Context: when a project group is scheduled to be deleted, we add it to a map<project, disposable> (like a queue). When the user opens a book right after they deleted a project, we want to "fast-forward" the deletion by disposing all the projects stored in the map and start deleting them immediately.

#### This PR makes sure once a project's delete() enters its "critical section", it must be FIRST removed from the queue before executing the delete task. Otherwise, the code would accidentally dispose() the deletion during its critical section and throw an exception.

> projectsWithDeleteTimer.remove(cardModel) // must call before executing actual delete()

This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1101)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1101#-:-NueP5FXCfvZOAi5XCEH:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #1148, opened 2024-05-28
Enables/disables peer edit confirm button on undo/redo (OT-890)

darrellcolehill:
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1148)


AnonymousWalker:
Let's not assume the view model has only one undoable action. Things like this is difficult to keep track if we later have additional actions supported.

See https://github.com/Bible-Translation-Tools/Orature/pull/1149

PR #1165, opened 2024-07-19
Exclude non-verse markers when highlighting

AnonymousWalker:
Source text are split by line break `\n`, thus it has no information about which marker each text item is associated with.
This PR excludes the Book and Chapter Marker(s) when computing the index to highlight - which only applies to Translation View and Edit Marker View. 

Narration is not affected, since the teleprompter items `NarratableItemModel` contains Chunk and AudioMarker
 

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1165)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1165#-:-O2BFyXt9iO6iUDgJ9B8:bnfp4nl)* status: 0 of 8 files reviewed, all discussions resolved

PR #1203, opened 2025-02-24
Wk v3.1.30 qa fix

wkelly17:
This PR fixes 2 unrelated bugs, and adds 2 small features.  It's not much code at all, but it was enough of a headache for me that I'm going to leave a record here of what left me confused for quite a while on this stuff. 

## Problem 1
1. The microphone permission prompt stopped working on fresh installs. 

Why: 
When we changed to signing the Mac app, install4j enables the hardened runtime. This is completely separate and different from sandboxing.  There are two similar but slightly different permissions for sandboxes vs the hardened runtime.  
See: https://stackoverflow.com/questions/49595811/macos-entitlements-audio-input-vs-microphone
In our case, when Dan had added com.apple.security.device.microphone, it worked, but now no longer did since singing enable the runtime. 

Solution: 
You'll notice there is now a 
entitlements-dmg.plist file and a entitlements.plist.   This is due to the fact that the sandbox needs some additional permissions, such as both audio entitlements and team identifier stuff related to signing.  

To Test: 
A fresh install of the normal mac dmg job should prompt you for mic permission. 

## Problem 2. 
Tests were passing, Code was successfully being notarized when you checked the cert chain and signature, BUT upon running Orature, app store builds crashed immediately.  

### Why: 
 - One of Orature's deps is a sqlite driver.   This sqlite driver works by extracting native drivers for platforms at runtime.  See https://github.com/xerial/sqlite-jdbc?tab=readme-ov-file#:~:text=An%20appropriate%20native%20library%20file%20is%20automatically%20extracted%20into%20your%20OS%E2%80%99s%20temporary%20folder%2C%20when%20your%20program%20loads%20org.sqlite.JDBC%20driver.
 - This behavior however, causes a problem for Macos sandbox apps, because it's trying to load code that has not been previously signed.  Hence you'd get an error along the lines of {hash}.tmp has not been signed.  

### Solution 
Mostly, just read this: https://groups.google.com/g/jna-users/c/Bws1h060faA?pli=1
- But, in a sentence, all lib code must be vendored in.  The is what the dylibs are that have now been placed in /nativelibs/mac.  
- JNI dispatch must match the version resolved by gradle dependnecies check, or it will throw.  This is currently 5.10 I think of JNI, but the tools it actually unpacks have a different version which I think is 6.1.2.   (the app will error if it's expecting one version and gets another).   Libsqlite there is just the latest and whatever is compatible with Orature. 
- Vendoring in these deps is a normal use case of the sqlite driver linked above via some jvm args.  And then JNA/I also can be configured to look in specific directories.  Altogether then, that's these args: 
`-Dorg.sqlite.lib.name=libsqlitejdbc.dylib,  -Djna.nounpack=true, -Djna.nosys=true, -Djna.boot.library.path=${installer:sys.contentDir}/mac/x86_64:${installer:sys.contentDir}/mac/aarch64` 
These are passed Only for the MacOs launcher for pkg files.  Every other target is still the same. 

## Feature
1. In the dedicated mac pkg build, I added a workflow parameter that allow you to build a dev version of the pkg file.  You can't install pkg's outside of app store if they are signed with distribution certificates, so you have to sign with a different set of keys for local testing of a pkg file (i.e. this cert has device ids for Joe, Max, and David).  This is slightly less of an issue because feature 2

2. You'll see that the install4j version bumped to 11.02.  This fixes a proble that was being had where install4j uses the VERSION for the id that apple uses to bump versions in the app store.   Now you can keep the marketing version the same, and it uses a timestamp under the hood so that you can test builds through testflight without incrementing the software version number. 

3. Following Dan's change, bumped the ubuntu runner to 24.04. 

As for testing, David tested through tesflight on intel and Mac m1-4, and then it worked on my machine with dev profile as well as Joe's.  Everyone reported microphone permissions where needed. Testflight for latest build showed no crashes. 

Lmk if you have any questions

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1203)


danparisd (approved):
LGTM

AnonymousWalker:
Crash log:
> Caused by: org.sqlite.NativeLibraryNotFoundException: No native library found for os.name=Windows, os.arch=x86_64
...

We probably missed Windows. @wkelly17 Could you check this and perhaps Linux?

PR #1207, opened 2025-03-13
Wk rm jfx web

wkelly17:
Ok, here we go again. Time to cut a new release, and I sure hope I'm to the end of breaking things.  For posterity sake, we have:

1. This error: 
image

To Fix: 
 - I thought upgrading controlsfx would do it since we used it as a dep and transitively javafx.web was a dep on our version, but this didn't fix this.  But it's good to upgrade anyway
 - The bundled jre (liberica 21) includes javafx (siince Orature isn't modularized.).  That Jre, when you look at the modules, has the following. 
image
- So, we told install4j to exclude   <module name="javafx.web" exclude="true" /> from all bundles (can't do this on a platform basis afaik.  JRE bundled is same per app. 
- This still didn't fully fix the issue, cause that jre also includes this dylib, which is what apple was mad about in the first place
- 
image
- i4j also let's you say broad exlcusion rules, hence the <files globalExcludeSuffixes="**/lib/libjfxwebkit.dylib">. Which, for the latest mac build shows it's no longer there. 
image

Next to last, Joe said we can kill tests for the mac build since it and the rest of the builds should stay in lockstep.  

Lastly, this already merged into dev, but for records keeping, I did this:https://github.com/Bible-Translation-Tools/Orature/blob/dev/jvm/workbookapp/otter.install4j#L927-L941, which is why the 3.1.30 linux and window builds bricked. I forgot / didn't full realize i4j for each media was building each launder, so that exclude for those limits them to the previous launcher.  The mac dmg launcher is fine cause you can only have one on apple platforms anyway.  

Here's to hoping it passes review and no more breakage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1207)

### DIFF
--- a/.github/workflows/mac_store_build.yml
+++ b/.github/workflows/mac_store_build.yml
@@ -130,7 +130,7 @@ jobs:
           -e MAC_NOTARY_ISSUER \
           -e MAC_NOTARY_KEY_ID \
           wycliffeassociates/install4j-docker:11.0.2 \
-          ./gradlew build :jvm:workbookapp:install4jdeploy "-Dorg.gradle.jvmargs=-Xmx4096m -Dnet.bytebuddy.experimental=true -XX:MaxMetaspaceSize=1024m" -x test
+          ./gradlew build :jvm:workbookapp:install4jdeploy "-Dorg.gradle.jvmargs=-Xmx4096m -Dnet.bytebuddy.experimental=true -XX:MaxMetaspaceSize=1024m" -x test -x integrationTest
         env:
           UPDATE_PATH: ${{ needs.set-env-vars.outputs.UPDATE_PATH }}
           ORG_GRADLE_PROJECT_gradlewCommandVersionProp: ${{ needs.set-env-vars.outputs.GHA_VERSION }}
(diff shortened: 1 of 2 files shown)


danparisd:
Are we using the most up to date Liberica JRE? If not, could we update it and maybe that would fix it?

AnonymousWalker:
> Are we using the most up to date Liberica JRE? If not, could we update it and maybe that would fix it?

I think so for JDK 21
https://github.com/Bible-Translation-Tools/Orature/blob/dev/jvm/workbookapp/otter.install4j#L957

jsarabia:
@danparisd if you mean that updating Liberica would fix Apple's issue with webkit, no, we looked at the latest version of javafx and it still has the issue, so Liberica would have the issue as well.

If we're not using it, and it saves 100mb in the binaries though, it's a huge win

jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1207#-:-OLJrb4w-I7BY1OwY3Ob:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @AnonymousWalker and @danparisd)

danparisd:
> @danparisd if you mean that updating Liberica would fix Apple's issue with webkit, no, we looked at the latest version of javafx and it still has the issue, so Liberica would have the issue as well.
> 
> If we're not using it, and it saves 100mb in the binaries though, it's a huge win

Sounds good to me!

PR #1208, opened 2025-03-20
Fix App Store Install breaking on Intel Arch

wkelly17:
Removes trying to pass org.lib.sqlite.path as a vm parameter and instead sets it in app code before sqlite jdbc driver code is loaded via that companion init block.  Only affects .pkg macs due to checking for a jvm arg passed through the i4j launcher.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1208)


jsarabia (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/Bible-Translation-Tools/Orature/1208#-:-OLp1nk7C3H1N7MLdRFo:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved

